PR #84, opened 2021-10-21
Update: Add  trigger and memory card

dani69654:
---
name: 'Pull request'
about: Add `callPluginEndpoint` trigger and memory card to temporary store data returned by the plugin
---

### Checklist

### Issue Description

It is necessary a way to chain multiple triggers (A | B ) where B depends by the result of A.
The logic allows to do so, temporary saving the result of A and use it to call B.
The plugin must return the right settings so that the trigger loop does not end until the chain is completed.

### Solution Description

Add `callPluginEndpoint` trigger.
Add new `MemoryCard` class.

### DIFF
--- a/src/MoralisWeb3.js
+++ b/src/MoralisWeb3.js
@@ -14,6 +14,9 @@ import detectEthereumProvider from '@metamask/detect-provider';
 import createSigningData from './createSigningData';
 const EventEmitter = require('events');
 const transferEvents = new EventEmitter();
+import { MemoryCard } from './MoralisWeb3MemoryCard';
+
+const memoryCard = new MemoryCard();
 
 export const EthereumEvents = {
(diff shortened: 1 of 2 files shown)


src/MoralisWeb3.js
```diff
@@ -14,6 +14,9 @@ import detectEthereumProvider from '@metamask/detect-provider';
 import createSigningData from './createSigningData';
 const EventEmitter = require('events');
 const transferEvents = new EventEmitter();
+import { MemoryCard } from './MoralisWeb3MemoryCard';
+
+const memoryCard = new MemoryCard();
```
ErnoW, line 19: It makes more sense to me to have the `memoryCard` as a static property in the `MoralisWeb3` class.

PR #117, opened 2021-11-12
BUGFIX: missing query params when using api key

rphansen91:
---
Options are not sent to api
---

Passing a chain to the Moralis.Web3Api always sent results for ETH

### Checklist

- [X ] I am not disclosing a [vulnerability](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/main/SECURITY.md).
- [ X] My code is conform the [code style](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/main/CODE_STYLE.md)
- [ X] I have made corresponding changes to the documentation
- [ X] I have updated Typescript definitions when needed

### Issue Description

The `axios.create` method of axios does not pass the params in the query string leading to inaccurate results.

### Solution Description

Use the main axios api to ensure the query params are sent along with the request

### DIFF
--- a/scripts/genWeb3API.js
+++ b/scripts/genWeb3API.js
@@ -108,16 +108,15 @@ static async fetch({ endpoint, params }) {
   try {
     const parameterizedUrl = this.getParameterizedUrl(url, params);
     const body = this.getBody(params, bodyParams);
-    const http = axios.create({
-      baseURL: this.baseURL,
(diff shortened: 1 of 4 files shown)


sogunshola (approved):
Looks good to me, I will also make this adjustment in the code generator file so subsequent builds have this update.

PR #125, opened 2021-11-19
Improve authenticate() function

mobilekosmos:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

1. user.setACL() is being called before checking if user is undefined/null
2. ParseObject already has a function for adding objects while avoiding duplicates:
that's why I replaced `user.set(...)` by `user.addAllUnique('accounts', accountsLower);`
There are other parts in the file where the own function `uniq(...)` is used, so I think if my change is accepted you won't really need that function. I didn't test this change in the JS-SDK itself though, I tested it separately, so I hope you have tests in place to validate the change.


ErnoW:
Looks good, thanks for the improvements and suggestions 👍

ErnoW (approved):

PR #127, opened 2021-11-21
feature: check if sdk is updated

dani69654:
---
name: 'Pull request'
about: Notify the user when the SDK is not updated
---

### Checklist

- [] I have made corresponding changes to the documentation
- [] I have updated Typescript definitions when needed

### Issue Description
Need a way to notify users when the SDK is not updated.

### Solution Description
Match the version in pkg.json with the one returned by the npm registry.
Could not use a better way because browserify breaks if we import package.json

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,14 +1,14 @@
-import axios from 'axios';
 import RESTController from './RESTController';
 
 const DEEP_INDEX_API_HOST = 'deep-index.moralis.io';
 const DEEP_INDEX_SWAGGER_PATH = '/api-docs/v2/swagger.json';
 
 const fetchSwaggerJson = async () => {
-  const http = await axios.create({ baseURL: `https://${DEEP_INDEX_API_HOST}` });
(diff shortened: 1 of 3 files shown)


src/utils.js
```diff
@@ -54,8 +55,27 @@ const fetchEndpoints = async () => {
...
+    );
+    const latestVersion = response.objects[0].package.version;
+    const installedVersion = process.env.npm_package_version;
+    if (installedVersion < latestVersion)
+      // eslint-disable-next-line no-console
+      console.warn(
+        'You are not using the latest version of the SDK. Please update it as soon as possible to enjoy the newest features.'
+      );
+  } catch (error) {
+    throw new Error('Could not verify SDK version');
```
ErnoW, line 72: Let's not throw an error but show a console warn here.
If npm is down, then we don't want all moralis apps to crash on startup
```suggestion
    // eslint-disable-next-line no-console
    console.warn('Could not verify SDK version');
```

ErnoW (changes requested):
Nice work, one request

ErnoW (approved):

PR #128, opened 2021-11-24
Update UnitConvert.js

Y0moo:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

Standard BigNumber utils from Web3.js don't properly support floats.

### Solution Description

We can use this solution or add a library bignumber.js

### DIFF
--- a/src/UnitConvert.js
+++ b/src/UnitConvert.js
@@ -9,7 +9,7 @@ class UnitConverter {
   }
 
   static Token(value, decimals) {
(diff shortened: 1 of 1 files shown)


ErnoW:
@Y0moo I'm not sure what the issue is, as tokens are always whole numbers. What is the use-case of supporting floats?
We will switch to ethers.js in the SDK, which has BN.js as library, so we will use that one for these functions.

ErnoW:
Not needed anymore in v1.0 with EthersJs support

PR #138, opened 2021-12-03
Feature/feature/fantom authentication for solana

ayorcodes:
---
name: 'Phantom authentication for solana'
---

### Checklist

### Issue Description

`Moralis.authenticate` does not support phantom authentication.

### Solution Description

Added support for Phantom authentication to `Moralis.authenticate`. 
Can be accessed by doing the following;

```
await Moralis.authenticate({
  type: 'phantom',
  signingMessage: 'Lorem ipsums',
});
  ```

### CODE REVIEW

File: src/MoralisPhantom.js [hunk 1]
```diff
@@ -0,0 +1,95 @@
...
+
+class MoralisPhantom {
+  static getProvider = () => {
+    if ('solana' in window) {
+      const provider = window.solana;
+      if (provider.isPhantom) {
+        return provider;
+      }
+    }
+    window.open('https://phantom.app/', '_blank');
```
ErnoW, line 17: I'm not a fan of opening the phantom website, if the user has the app not installed, lets leave this implementation to the devs, if they want to do this.

src/MoralisPhantom.js
```diff
@@ -0,0 +1,95 @@
...
+// const { clusterApiUrl, Connection } = require('@solana/web3.js');
+import createSigningData from './createSigningData';
+import ParseUser from './ParseUser';
+import ParseQuery from './ParseQuery';
+import ParseObject from './ParseObject';
+import ParseACL from './ParseACL';
+
+class MoralisPhantom {
+  static getProvider = () => {
+    if ('solana' in window) {
```
ErnoW, line 11: Please also check if window is defined, to avoid errors in non-browser environments

src/MoralisPhantom.js
```diff
@@ -0,0 +1,95 @@
+/* global window */
+// const { clusterApiUrl, Connection } = require('@solana/web3.js');
+import createSigningData from './createSigningData';
+import ParseUser from './ParseUser';
+import ParseQuery from './ParseQuery';
+import ParseObject from './ParseObject';
+import ParseACL from './ParseACL';
+
+class MoralisPhantom {
+  static getProvider = () => {
```
ErnoW, line 10: We can call this `enable` to ensure consistency. And thow the `Phantom wallet not available` error if the provider could not be found.
Then also move the `connect` call in here.
That way it is most similar to web3 implementation

ErnoW (changes requested):

ErnoW (approved):

PR #139, opened 2021-12-06
fix: cors error on version check

ErnoW:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

Requesting version info from npm results in a  failing preflight check on firefox and results in a CORS issue.

### Solution Description

- Using unpkg as domain to fetch the latest version
- Updated the check to compare versions to account for any semantic versioning

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -78,7 +78,7 @@ const checkForSdkUpdates = async () => {
   try {
     const { response } = await RESTController.ajax(
       'GET',
-      'https://www.unpkg.com/moralis@0.0.148/package.json'
+      'https://www.unpkg.com/moralis/package.json'
     );
     const latestVersion = response.version;
     const installedVersion = process.env.npm_package_version;
--- a/src/utils.js
+++ b/src/utils.js
@@ -55,22 +55,41 @@ const fetchEndpoints = async () => {
   return data;
 };
 
+/**
+ * Compares if the semantic version of version1 is larget than version2
+ */
+const isSemanticVersionLarger = (version1, version2) => {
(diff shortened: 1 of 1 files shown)


src/utils.js
```diff
@@ -55,22 +55,41 @@ const fetchEndpoints = async () => {
...
+
+  return false;
+};
+
 const checkForSdkUpdates = async () => {
   try {
     const { response } = await RESTController.ajax(
       'GET',
-      'https://registry.npmjs.org/-/v1/search?text=moralis&size=1'
+      'https://www.unpkg.com/moralis@0.0.148/package.json'
```
dani69654, line 81: It seems that GET `https://www.unpkg.com/moralis@0.0.148/package.json` returns info about that specific version, not the latest one

src/utils.js
```diff
@@ -55,22 +55,41 @@ const fetchEndpoints = async () => {
   return data;
 };
 
+/**
+ * Compares if the semantic version of version1 is larget than version2
+ */
+const isSemanticVersionLarger = (version1, version2) => {
```
dani69654, line 61: Is this function needed? JS should be able to compare two floating point numbers

src/utils.js
ErnoW: Ah you're right. Removed the version. Now it automatically resolves to the most recent version

src/utils.js
ErnoW: Yes currently that would work since our versions are `0.0.xxx`. This makes it future-proof when we integrate semantic versioning. For example "0.0.39" < "0.0.4" if we compare it as a string.

PR #141, opened 2021-12-08
remove node warning in moralis.start

sogunshola:
---
name: 'Pull request'
about: Remove info message when using node without moralis secret
---

### Checklist

### Issue Description

### DIFF
--- a/src/Parse.js
+++ b/src/Parse.js
@@ -52,7 +52,7 @@ class Moralis extends MoralisWeb3 {
     } else {
       // Node environment
       if (!moralisSecret) {
-        console.warn('Moralis.start warning: to use web3 access, moralisSecret is required');
+        // console.warn('Moralis.start warning: to have web3 access, moralisSecret is required');
       } else {
         this.moralisSecret = moralisSecret;
         const { web3ApiKey, speedyNodeApiKey } = await this.getApiKeys(moralisSecret);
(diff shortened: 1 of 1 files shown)


src/Parse.js
```diff
@@ -52,7 +52,7 @@ class Moralis extends MoralisWeb3 {
     } else {
       // Node environment
       if (!moralisSecret) {
-        console.warn('Moralis.start warning: to use web3 access, moralisSecret is required');
+        // console.warn('Moralis.start warning: to have web3 access, moralisSecret is required');
```
ErnoW, line 55: Remove the code instead of commenting it out.

Also you can convert the `if-else` to an `if` check

ErnoW (changes requested):

src/Parse.js
sogunshola: 👍🏽 great suggestion.

https://github.com/MoralisWeb3/Moralis-JS-SDK/pull/141/commits/6b29197b7297f780ed821802932a978811d4b2a0

done that

src/Parse.js
```diff
@@ -52,8 +52,6 @@ class Moralis extends MoralisWeb3 {
     } else {
       // Node environment
       if (!moralisSecret) {
```
ErnoW, line 54: Reverse this check please as it is incorrect now

```suggestion
      if (moralisSecret) {
```

ErnoW (changes requested):

ErnoW (approved):

Issue #168, opened 2021-12-31
[Bug] Custom Enable Web3 Persistence (Can't Open Metamask or WalletConnect after setEnableWeb3)


YosephKS:
### Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/main/SECURITY.md).
- [x] I have searched through [existing issues](https://github.com/MoralisWeb3/Moralis-JS-SDK/issues?q=is%3Aissue) and the [Moralis Forum](https://forum.moralis.io/).
- [x] I can reproduce the issue with the [latest SDK release](https://github.com/MoralisWeb3/Moralis-JS-SDK/releases), and my server is updated to the latest version.

### Issue Description

Unable to open Metamask or WalletConnect after `Moralis.setEnableWeb3`. This is a problem because user might cancel connecting with custom provider and instead choose Matamask/WalletConnect to connect to the dApp instead. Based on the code `src/MoralisWeb3.js`, seems like it will store the custom enable web3 function in `this.customEnableWeb3` here:

```js
static setEnableWeb3(fn) {
    this.customEnableWeb3 = fn;
}
```

which then when trying to `Moralis.authenticate` or `Moralis.enableWeb3`, will first check whether there is `this.customEnableWeb3` first in the if-else logic as the code showed below.

```js
static async enableWeb3(options) {
    let web3;

    if (this.customEnableWeb3) {
      web3 = await this.customEnableWeb3(options);
    } else {
      if (this.speedyNodeApiKey) {
        options.speedyNodeApiKey = this.speedyNodeApiKey;
        options.provider = 'custom';
      }
      const Web3Provider = MoralisWeb3.getWeb3Provider(options);
      const web3Provider = new Web3Provider();
      this.activeWeb3Provider = web3Provider;

      web3 = await web3Provider.activate(options);
    }

    this.web3 = web3;
    return web3;
  }
```

Since `this.customEnableWeb3` persisted, whenever the user try to run `Moralis.authenticate` or Moralis.enableWeb3` it will instead call the custom wallet provider instead of Metamask/WalletConnect.

### Steps + code to reproduce

1. Add `moralis` + `react-moralis` to create-react-app
2. Wrap the `<App />` component with `MoralisProvider` and add `appId` and `serverUrl`
3. Add multiple options for logging in with Moralis Authentication, like this
```ts
/**
   * @description Handle Wallet authentication
   *
[...]

PR #178, opened 2022-01-05
Fix Custom Web3 Wallet Authentication with Moralis SDK

YosephKS:
---
name: Fix Custom Enable Web3 for Moralis Authentication/Enable Web3
about: Cannot authenticate with Metamask/WalletConnect after setEnableWeb3
---

### Checklist

### Issue Description

After `setEnableWeb3` is called to set custom web3 wallet provider, authentication with Metamask/Walletconnect can't be triggered. Instead, it keeps persisting calling the custom web3 wallet.

Related issue: https://github.com/MoralisWeb3/Moralis-JS-SDK/issues/168

### Solution Description

Delete `setEnableWeb3` and add `customEnableWeb3` field as a replacement in the `options` field for `Moralis.authenticate` and `Moralis.enableWeb3`

### DIFF
--- a/src/MoralisWeb3.js
+++ b/src/MoralisWeb3.js
@@ -39,15 +39,11 @@ class MoralisWeb3 {
     return this.ensureWeb3IsInstalled();
   }
 
-  static setEnableWeb3(fn) {
-    this.customEnableWeb3 = fn;
-  }
-
   static async enableWeb3(options) {
     let web3;
 
-    if (this.customEnableWeb3) {
-      web3 = await this.customEnableWeb3(options);
+    if (options?.provider === 'customWallet' && options?.customEnableWeb3) {
+      web3 = await options.customEnableWeb3();
     } else {
       if (this.speedyNodeApiKey) {
         options.speedyNodeApiKey = this.speedyNodeApiKey;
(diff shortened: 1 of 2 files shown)


YosephKS:
@ErnoW not sure about the solution I have here, let me know what are your thoughts.

Also best not to merge after approving, might need to make a bit of similar changes in `react-moralis` and the documentation. 🙌🏼

ErnoW:
> @ErnoW not sure about the solution I have here, let me know what are your thoughts.
> 
> Also best not to merge after approving, might need to make a bit of similar changes in `react-moralis` and the documentation. 🙌🏼

I think this will not be needed anymore, as we have a different approach, currently in beta (the ethers refactor).
With that approach, the user needs to provide a `connector` (custom implementation of connecting to web3) along with the authenticate/enableWeb3 functions. This means that the functions will not be replaced, and remain intact for future calls.

YosephKS:
> > @ErnoW not sure about the solution I have here, let me know what are your thoughts.
> > Also best not to merge after approving, might need to make a bit of similar changes in `react-moralis` and the documentation. 🙌🏼
> 
> I think this will not be needed anymore, as we have a different approach, currently in beta (the ethers refactor). With that approach, the user needs to provide a `connector` (custom implementation of connecting to web3) along with the authenticate/enableWeb3 functions. This means that the functions will not be replaced, and remain intact for future calls.

Ahhh I see, got it good to hear! Alright I will close this PR then~

PR #235, opened 2022-01-28
Enable passing argument to `resolveAddress`

JackCurrie:
- Fix type signature for `resolver:resolveAddress` method on the
  `Web3Api` class such that it can take in its intended options

---
name: Enable passing argument to `resolveAddress`
about: Reverse Resolution from Eth address to ENS names
---

### Checklist

### Issue Description

As a user, I was attempting to use the `resolve/:address/reverse` API via the `Web3Api > resolve > resolveAddress` method in my typescript project, and it was not working. The type signature doesn't allow me to pass in options to this method with the address string, so I am unable to use this method.

Related issue: I have not filed an issue related to this, as the solution seems very straightforward. Please let me know if there's anything I can do to make the reviewer's life easier!

### Solution Description

The solution is very simple--just a one-line adjustment to the type signature for this method. Very small and straight forward, and I don't believe any other adjustments need to be made.

### DIFF
--- a/types/generated/web3Api.d.ts
+++ b/types/generated/web3Api.d.ts
@@ -2558,7 +2558,7 @@ export default class Web3Api {
 
   static resolve: {
     resolveDomain: (options: operations["resolveDomain"]["parameters"]["query"] & operations["resolveDomain"]["parameters"]["path"]) => Promise<operations["resolveDomain"]["responses"]["200"]["content"]["application/json"]>;
-    resolveAddress: () => Promise<operations["resolveAddress"]["responses"]["200"]["content"]["application/json"]>;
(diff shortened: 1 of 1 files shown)


ErnoW:
Thanks for brining this up, but the file is auto-generated.
I will check the generation script, as I think we already fixed a similar issue for the solanaAPI

JackCurrie:
@ErnoW thanks for the feedback. I'm a bit of a typescript rookie--I saw that it was in a `generated` directory, but couldn't figure out where the generated code came from...

I'll poke around a bit and see if I can resolve this as well.

ErnoW:
@JackCurrie 
No problem, it's not as obvious if you're not familiar with the setup.

Fixed it in https://github.com/MoralisWeb3/Moralis-JS-SDK/pull/236

Released in v1.2.3

JackCurrie:
@ErnoW thank you sir! super helpful, was just digging into the generation script, this saved me alot of time. 😄

PR #243, opened 2022-02-01
Feat/web3 auth

rayyan224:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description


src/Web3Connector/Web3AuthConnector.js
```diff
@@ -0,0 +1,102 @@
...
+    }
+    if (!displayName) {
+      throw new Error('"displayName" not provided, please provide displayName');
+    }
+    if (!clientId) {
+      throw new Error('"clientId" not provided, please provide clientId');
+    }
+
+    // Initalizing Web3Auth
+    const { Web3Auth } = require('@web3auth/web3auth');
```
kantorcodes, line 23: Why are we using require statements? 😟 😓

src/Web3Connector/Web3AuthConnector.js
```diff
@@ -0,0 +1,102 @@
...
+    const { CHAIN_NAMESPACES, ADAPTER_STATUS } = require('@web3auth/base');
```
kantorcodes, line 24: Why are we using require statements? 😟 😓

kantorcodes: This is really bad practice for keeping bundle sizes small

ErnoW: At the moment there are no real benefits that we can get, as the SDK is not set up to benefit from tree-shaking. Also in this case we need to load the dependency dynamically (as it's an optional dependency), so using require is logical to use here.

That said, I understand your concern regarding bundle size, and we will work on this in the next major version of the SDK, to make it more modular by using separate packages, and rollup (and probably other tech improvements).

kantorcodes: It's possible to benefit from dynamic imports with gulp and babel, see

 https://link.medium.com/ffwfIIIMjnb

Has this approach been tried?

By using require you create tech debt and make tree shaking more and more difficult for each new require statement that is written. Furthermore, I don't believe the variant used in these lines of code is actually dynamic. This is a misconception.

You can use require.ensure to create a chunk instead. Unless you mean the node module is omitted entirely from the bundle during compile process? ( I have no idea how you would reasonably keep track of that, but given the current bundle size I don't believe it's going well.)

https://www.jianshu.com/p/b2dcf4d0e78e

ErnoW: Yes we tried a similar approach, but without the webpack setup.

But yes, currently the module is omitted from the compiled sdk, and should only be added when the user wants to include this functionality (similar to magic-sdk and walletconnect). But like I mentioned, we will make it its own package in the next release.

Anyway, thanks for the tips and references 🙏 , we will look into it if it is possible with a similar setup by using webpack.

kantorcodes: Awesome thank you. 

Another tip, with modern browsers, axios should not be needed, and `fetch` can safely replace all API calls.

src/MoralisWeb3.js
```diff
@@ -195,6 +196,8 @@ class MoralisWeb3 {
         return NetworkWeb3Connector;
       case 'magicLink':
         return MagicWeb3Connector;
+      case 'Web3AuthConnector':
```
ErnoW, line 199: lets rename this to web3Auth

src/Web3Connector/Web3AuthConnector.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+    if (!displayName) {
+      throw new Error('"displayName" not provided, please provide displayName');
+    }
+    if (!clientId) {
+      throw new Error('"clientId" not provided, please provide clientId');
+    }
+
+    // Initalizing Web3Auth
+    const { Web3Auth } = require('@web3auth/web3auth');
```
ErnoW, line 23: Similar to walletconnect and magicLink,
This require statement needs to be dynamic. And the `@web3auth/web3auth` dependency needs to be optional. So:
- Wrap it in a try/catch
- Test if you get no error when you build the app and have the dependency not included
- Add this module to `filteredWarnings` in `src/filterConsole.js`

Currently I get an error when building an app without the dependencies. We only want an error to show if the dev actually uses this connector without having the dependency installed. But it should be all optional.

src/Web3Connector/Web3AuthConnector.js
```diff
@@ -0,0 +1,104 @@
...
+    const { CHAIN_NAMESPACES, ADAPTER_STATUS } = require('@web3auth/base');
```
ErnoW, line 24: See above comment, same holds for this dependency

src/Web3Connector/Web3AuthConnector.js
```diff
@@ -0,0 +1,104 @@
...
+    if (!Web3Auth) {
```
ErnoW, line 25: Also, we had to include something like:
```
WalletConnectProvider = window?.WalletConnectProvider?.default
```
for walletconnect. This was needed to make it work for non-npm-builds, for example if the users imports moralis and this library via a `<script/>` from a CDN provider.
I guess that something similar is also needed in this case.

Please test if it works in a simple app by only using html/js and by using CDN imports via script tags

src/Web3Connector/Web3AuthConnector.js
```diff
@@ -0,0 +1,104 @@
...
+    if (!ADAPTER_STATUS) {
+      throw new Error('"@web3auth/base" not installed, please install');
+    }
+    const ethChainConfig = {
+      chainNamespace: CHAIN_NAMESPACES.EIP155,
+      chainId: hexChainId,
+      rpcTarget: rpcTarget,
+      displayName: displayName,
+      blockExplorer: 'https://ropsten.etherscan.io/',
+      ticker: 'ETH',
```
ErnoW, line 40: Does this mean that the ticker is always ETH?
So for example also if I use Binance Smart Chain? In that case I would expect it to be BNB.
Would be nice if `web3auth` could determine that on its own. Othewise we would need a mapping for this,

src/Web3Connector/Web3AuthConnector.js
```diff
@@ -0,0 +1,104 @@
...
+import AbstractWeb3Connector from './AbstractWeb3Connector';
+
+export class Web3AuthConnector extends AbstractWeb3Connector {
+  type = 'Web3AuthConnector';
+
+  activate = async ({ rpcTarget, hexChainId, displayName, clientId }) => {
+    // Checking that all params are given
+    if (!rpcTarget) {
+      throw new Error('"rpcUrl" not provided, please provide infuraId');
+    }
```
ErnoW, line 11: We don't need the user to provide an infureId link. We can use our own rpc-urls. We can re-use the rpc's from walletconnect in this case:
```
getMoralisRpcs('WalletConnect')[chainIdInDecimals]
```

src/Web3Connector/Web3AuthConnector.js
```diff
@@ -0,0 +1,104 @@
...
+    if (!hexChainId) {
```
ErnoW, line 12: Also let the user inut chainId in decimal form.
We can use the `verifyChainId(providedChainId)` for this. This will transform number chainIds to hex

src/Web3Connector/Web3AuthConnector.js
```diff
@@ -0,0 +1,104 @@
...
+      throw new Error('"hexChainId" not provided, please provide hexChainId');
+    }
+    if (!displayName) {
+      throw new Error('"displayName" not provided, please provide displayName');
+    }
```
ErnoW, line 17: The user should not worry about the displayname, only about the chainId.
Would be nice if web3Auth can infer the displayname automatically.
Otherwise we will need to make a mapping for this

src/Web3Connector/Web3AuthConnector.js
```diff
@@ -0,0 +1,104 @@
...
+    // Gathering data
+    try {
+      const isSocialLogin = web3auth?.provider ? false : true;
+      const ether = new ethers.providers.Web3Provider(
+        web3auth?.provider ? web3auth.provider : web3auth
+      );
+
+      const signer = ether.getSigner();
+      const { chainId } = await ether.getNetwork();
+      const address = (await signer.getAddress()).toLowerCase();
```
ErnoW, line 78: These requests can be in parrallel by using `Promise.all([])`

src/Web3Connector/Web3AuthConnector.js
```diff
@@ -0,0 +1,104 @@
+import { ethers } from 'ethers';
+import AbstractWeb3Connector from './AbstractWeb3Connector';
+
+export class Web3AuthConnector extends AbstractWeb3Connector {
+  type = 'Web3AuthConnector';
```
ErnoW, line 5: Rename to web3Auth (no need to add "Connector")

types/index.d.ts
```diff
@@ -331,7 +331,13 @@ export namespace Moralis {
...
   }
 
-  type Web3ProviderType = 'metamask' | 'walletconnect' | 'walletConnect' | 'wc' | 'magicLink';
+  type Web3ProviderType =
+    | 'metamask'
+    | 'walletconnect'
+    | 'walletConnect'
+    | 'wc'
+    | 'magicLink'
+    | 'Web3AuthConnector';
```
ErnoW, line 340: lets rename this to web3Auth

ErnoW (changes requested):

types/index.d.ts
rayyan224: sounds good!

src/Web3Connector/Web3AuthConnector.js
rayyan224: Yea I like the way u think!

ErnoW (approved):

PR #247, opened 2022-02-03
added functionality to use overload contract methods

matiyin:
added functionality to use overload contract methods such as safeTransferFrom in ERC721

---
name: 'Suggested fix to use overload functions'
about: overload contract functions not working in executeFunction
---

### Checklist

### Issue Description

Cannot use safeTransferFrom overloaded function with Moralis.executeFunction, see:
https://forum.moralis.io/t/cannot-use-safetransferfrom-overloaded-function-with-moralis-executefunction/8914

### Solution Description

I've added an extra flag to support using an overload function syntax, described in more detail in the above forum post.

### DIFF
--- a/src/MoralisWeb3.js
+++ b/src/MoralisWeb3.js
@@ -652,6 +652,7 @@ class MoralisWeb3 {
     contractAddress,
     abi,
     functionName,
+    functionOverload,
     msgValue,
     params = {},
     overrides = {},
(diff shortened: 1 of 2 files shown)


src/MoralisWeb3.js
```diff
@@ -699,7 +700,13 @@ class MoralisWeb3 {
     });
 
     const contract = new ethers.Contract(contractAddress, abi, signerOrProvider);
-
+    if (functionOverload) {
+      const types = [];
+      for (const input of functionData.inputs) {
```
ErnoW, line 705: Don't think that this logic will work.
As there are multiple entries in the abi, in this case we always use the first one that is found, and use that as input

ErnoW (commented):
I see the issue, seems like we need to handle it a bit differently for overloaded functions.

Don't think that this approach will work in all cases, as we still use the the functionData of the first entry in the abi.

Let me test it a little bit and see if I can find a nice solution to make it work with EthersJs

PR #256, opened 2022-02-09
fix: overloaded functions for executeFunction

ErnoW:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

Overloaded functions do not work

Related issue: https://github.com/MoralisWeb3/Moralis-JS-SDK/pull/247

### Solution Description

Functions that have overloaded function definitions should be called with a topic `getMessage(string)`

If an abi has multiple definitions, and no topic is specified in the functionName; an error will be thrown
If no abi definition can be found for the provided functionName+topic, an error will be thrown

If the functionName+topic is valid, then the correct functionData will be selected from the abi, validation checks will run and the contract is called with `functionName(topic)`, as expected by EthersJs

### DIFF
--- a/src/MoralisWeb3.js
+++ b/src/MoralisWeb3.js
@@ -680,9 +680,64 @@ class MoralisWeb3 {
       signerOrProvider,
     } = this.getInternalWeb3Provider();
 
-    const functionData = abi.find(x => x.name === functionName);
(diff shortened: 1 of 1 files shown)


matiyin:
excellect work @ErnoW, intuitive and elegant. Tested, thanks so much.

PR #279, opened 2022-03-02
added endpoint to track and implemented in start function

sogunshola:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

### DIFF
--- a/src/Parse.js
+++ b/src/Parse.js
@@ -38,6 +38,7 @@ class Moralis extends MoralisWeb3 {
   static async start(options) {
     const { appId, serverUrl, plugins, javascriptKey, masterKey, moralisSecret } = options;
     let apiKey;
+    let trackOptions;
 
     if (process.env.PARSE_BUILD !== 'node') {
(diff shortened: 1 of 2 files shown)


ErnoW (changes requested):

src/Parse.js
```diff
@@ -68,11 +69,31 @@ class Moralis extends MoralisWeb3 {
     this.Web3API.initialize({ serverUrl, apiKey, Moralis });
     this.SolanaAPI.initialize({ serverUrl, apiKey, Moralis });
     if (appId && serverUrl) {
+      trackOptions = {
+        subdomain: serverUrl.split('/')[2].split(':')[0],
```
ErnoW, line 73: This can result in errors if the serverUrl is not given in the specific syntax that we need. Most likely this will not result in errors now, but what if we decide to change the serverUrl structure in the future.

Better is to make a function for example "getSubdomain(serverUrl)", and have checks in there to validate the input.

Also this function can be used later on in this file where we have the same logic.

PR #285, opened 2022-03-11
Optimisation/web3api fix

sogunshola:
---
name: 'Pull request'
about: When calling web3api the sessionToken is never sent forcing rate limit to always be anonymous. this PR handles that.
---

### Checklist

### Issue Description

### DIFF
--- a/scripts/genWeb3API.js
+++ b/scripts/genWeb3API.js
@@ -79,7 +79,25 @@ static getParameterizedUrl(url, params) {
   return parameterizedUrl;
 }
 
-static getErrorMessage(error, url) {
+static getApiRateLimitInfo(headers) {
+  return {
+    'x-rate-limit-limit': headers['x-rate-limit-limit'],
+    'x-rate-limit-remaining': headers['x-rate-limit-remaining'],
+    'x-rate-limit-remaining-ttl': headers['x-rate-limit-remaining-ttl'],
+    'x-rate-limit-throttle-limit': headers['x-rate-limit-throttle-limit'],
+    'x-rate-limit-throttle-remaining': headers['x-rate-limit-throttle-remaining'],
+    'x-rate-limit-throttle-remaining-ttl': headers['x-rate-limit-throttle-remaining-ttl'],
+    'x-rate-limit-throttle-used': headers['x-rate-limit-throttle-used'],
+    'x-rate-limit-used': headers['x-rate-limit-used'],
+    'x-request-weight': headers['x-request-weight'],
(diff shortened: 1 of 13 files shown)


scripts/genWeb3API.js
```diff
@@ -135,6 +136,19 @@ static async apiCall(name, options) {
     try {
       const http = axios.create({ baseURL: this.serverUrl });
       if (!options.chain) options.chain = 'eth';
+      if(process.env.PARSE_BUILD !== 'node') {
+        // eslint-disable-next-line no-undef
+        const sessionToken = JSON.parse(localStorage.getItem(\`Parse/\${this.appId}/currentUser\`));
```
ErnoW, line 141: We should not ignore the `no-undef` warning. It is possible that localStorage is not defined. For example react-native environment

Can we not use the Moralis or User object (after line 96)?

scripts/genWeb3API.js
```diff
@@ -135,6 +136,19 @@ static async apiCall(name, options) {
...
+        // eslint-disable-next-line no-undef
+        const installationId = localStorage.getItem(\`Parse/\${this.appId}/installationId\`);
```
ErnoW, line 143: Same as previous comment

scripts/genWeb3API.js
ErnoW: Otherwise we should add an extra check if `localStorage` is defined.
Always be as defensive as possible instead of ignoring warnings :)

ErnoW (changes requested):

scripts/genWeb3API.js
```diff
@@ -135,6 +136,19 @@ static async apiCall(name, options) {
...
+        if(sessionToken) {
+          options._SessionToken = sessionToken.sessionToken;
+        }
+        if(installationId) {
+          options._InstallationId = installationId;
+        }
+        options._ApplicationId = this.appId
```
ErnoW, line 150: Also, why is the installationId and _ApplicationId needed? In your solution you described we would need the sessionToken. Are these also needed for the rate-limiting?

sogunshola: Indeed it is not needed, I just tried to replicate what HttpRequest.request send, but it is not needed for rate limit. I will that out in the next push

scripts/genWeb3API.js
sogunshola: 👍🏾  I will make those adjustments now. thanks for pointing them out

ErnoW (approved):

PR #300, opened 2022-03-23
Web3auth Modal Open -> Close -> Open Bug

kallinlobus:
---
name: Web3Auth Modal Fix
about: Fixed an bug with opening and closing a modal in web3auth
---

### Checklist

### Issue Description
When using web3Auth for authentication an error is thrown when a user opens the web3auth modal, closes the modal without logging in, and then tries to open the modal again.

The error is: Error: Cannot execute Moralis.enableWeb3(), as Moralis Moralis.enableWeb3() already has been called, but is not finished yet

Related issue: https://forum.moralis.io/t/how-to-terminate-isauthenticating-or-isweb3enableloading/11681

### Solution Description
In Web3AuthConnector.js I added a listener to check the modal's status. If the modal is opened and then closed, a handleDisconnect error is sent back up the chain. The error message is 'Web3Auth: User closed login modal.'

In order for the error to be caught by the listener in InternalWeb3Provider.js, I had to move the initialization of the listeners there to before the "this.connector.activate(options);" call.

Lastly, I updated MoralisWeb3.js to handle the error by setting this.isEnablingWeb3 to false and then doing standard cleanup when a handleDisconnect error with message 'Web3Auth: User closed login modal.' is caught.

A node module with the updated code can be accessed here: https://www.npmjs.com/package/moralis-lobus

kallin@lobus.io is my email if you have questions/concerns. Thanks!


ErnoW:
Thanks for the well written PR.
Overall seems good to me. The integration with react-moralis still fails (authentication/enabling stuck in loading). 
Will do a few more tests and verify if this should be solved in the SDK or react-moralis.

Will come back to this after testing

ErnoW:
Fixed in https://github.com/MoralisWeb3/Moralis-JS-SDK/pull/314

PR #308, opened 2022-04-01
update log message if running on node.js

adamrhunter:
If using require('moralis') on node.js, the error message should provide the correct require command.  Updated to use moralis module (previously parse)

---
name: 'Pull request'
about: A new pull request
---

### Checklist

- [x ] I am not disclosing a [vulnerability](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/main/SECURITY.md).
- [ x] My code is conform the [code style](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/main/CODE_STYLE.md)
- [ x] I have made corresponding changes to the documentation
- [ x] I have updated Typescript definitions when needed

### Issue Description

The log message if the wrong version of the Moralis module is used with nodej.js provides the incorrect solution

### Solution Description

Updated the log message to include the correct require command.

### DIFF
--- a/src/Parse.js
+++ b/src/Parse.js
@@ -138,7 +138,7 @@ class Moralis extends MoralisWeb3 {
       /* eslint-disable no-console */
       console.log(
         "It looks like you're using the browser version of the SDK in a " +
-          "node.js environment. You should require('parse/node') instead."
+          "node.js environment. You should require('moralis/node') instead."
       );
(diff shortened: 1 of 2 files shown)


adamrhunter:
@ ErnoW - I fixed the test case that was failing as a result of the change to the message

ErnoW (approved):

PR #319, opened 2022-04-21
feat(api-pagination): Integrating pagination on web3api

sogunshola:
---
name: 'Web3api pagination'
about: This pull request implements pagination when using web3api seamlessly by calling `await response.next()`
---

### Checklist

### Issue Description

### DIFF
--- a/scripts/genWeb3API.js
+++ b/scripts/genWeb3API.js
@@ -79,6 +79,21 @@ static getParameterizedUrl(url, params) {
   return parameterizedUrl;
 }
 
+static getNextOptions(result, options) {
+  const nextOptions = options;
+  if (!result.page_size) return options
+  if (result.cursor && result.cursor !== undefined) {
+    if (nextOptions.cursor === result.cursor) return [];
+    nextOptions.cursor = result.cursor;
+  } else {
+    if (result.total > (result.page_size * result.page) + 1) {
+      nextOptions.offset = (result.page + 1) * (nextOptions.limit || 500);
+    }
+  }
+
+  return nextOptions;
+}
+
 static getApiRateLimitInfo(headers) {
   return {
     'x-rate-limit-limit': headers['x-rate-limit-limit'],
@@ -138,21 +153,9 @@ static async fetchFromApi(endpoint, params) {
       },
     });
     const result = response.data;
-    if(result.page_size) {
-      if(result.cursor) {
-        if(params.cursor === result.cursor) {
-          return [];
-        }
-        params.cursor = result.cursor;
-        result.next = () => this.fetchFromApi(endpoint, params);
-      }else {
-        if(result.total > (result.page_size * result.page) + 1){
-          params.offset = (result.page+1) * (params.limit || 500);
-          result.next = () => this.fetchFromApi(endpoint, params);
-        }
-      }
-    }
-      return result
+    const nextOptions = this.getNextOptions(result, params)
+    result.next = () => this.fetchFromApi(endpoint, nextOptions);
+    return result
   } catch (error) {
     const {status, headers, data} = error.response;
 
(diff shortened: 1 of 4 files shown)


scripts/genWeb3API.js
```diff
@@ -137,7 +137,22 @@ static async fetchFromApi(endpoint, params) {
         'x-api-key': this.apiKey,
       },
     });
-    return response.data;
+    const result = response.data;
+    if(result.page_size) {
+      if(result.cursor) {
+        if(params.cursor === result.cursor) {
```
Capplequoppe, line 143: Should we also have `&& params.cursor !== undefined`?

scripts/genWeb3API.js
```diff
@@ -137,7 +137,22 @@ static async fetchFromApi(endpoint, params) {
         'x-api-key': this.apiKey,
       },
     });
-    return response.data;
+    const result = response.data;
```
Capplequoppe, line 140: I think we can make this more DRY.
What about a helper function like this somewhere; 
```
function getNextOptions(result, options) {
  let nextOptions = options;
  const { result } = response.data;
  if (result.page_size) return options
  if (result.cursor) {
    if (nextOptions.cursor === result.cursor) return [];

    nextOptions.cursor = result.cursor;
  } else {
    if (result.total > (result.page_size * result.page) + 1) {
      nextOptions.offset = (result.page + 1) * (nextOptions.limit || 500);
    }
  }

  return options;
}
```

Then we can just assign the next function like this: 

```
const nextOptions = getNextOptions(result, options)
result.next = () => this.fetchFromServer(name, nextOptions);
```

or

```
const nextOptions = getNextOptions(result, options)
result.next = () => this.fetchFromApi(endpoint, nextOptions);
```

Capplequoppe (dismissed):

scripts/genWeb3API.js
```diff
@@ -79,6 +79,21 @@ static getParameterizedUrl(url, params) {
...
 }
 
+static getNextOptions(result, options) {
+  const nextOptions = options;
+  if (!result.page_size) return options
+  if (result.cursor && result.cursor !== undefined) {
+    if (nextOptions.cursor === result.cursor) return [];
+    nextOptions.cursor = result.cursor;
+  } else {
+    if (result.total > (result.page_size * result.page) + 1) {
```
ErnoW, line 89: Shouldn't this be: `result.total > (result.page_size * (result.page + 1))`?

ErnoW: We have no check if `result.total` or `result.page` is defined. This is probably the case when result.page_size is defined. But lets add the check to be safe

scripts/genWeb3API.js
```diff
@@ -79,6 +79,21 @@ static getParameterizedUrl(url, params) {
   return parameterizedUrl;
 }
 
+static getNextOptions(result, options) {
+  const nextOptions = options;
+  if (!result.page_size) return options
+  if (result.cursor && result.cursor !== undefined) {
```
ErnoW, line 85: No need to check for undefined if you check for `result.cursor`

scripts/genWeb3API.js
```diff
@@ -137,7 +152,10 @@ static async fetchFromApi(endpoint, params) {
         'x-api-key': this.apiKey,
       },
     });
-    return response.data;
+    const result = response.data;
+    const nextOptions = this.getNextOptions(result, params)
+    result.next = () => this.fetchFromApi(endpoint, nextOptions);
```
ErnoW, line 157: We don't always want to add a next function. We only want this for paginated requests (so when we use offset or cursor). Please add a conditional check for this.

ErnoW (changes requested):

scripts/genWeb3API.js
sogunshola: I agree, but it this was a suggestion from Niclas.

ErnoW:
Seems to be a use-case where the next() function is added, when it should not. See screenshot:
Screenshot 2022-04-25 at 13 55 29

Also
- Check the linting error
- add the next() function as optional Typescript type

ErnoW:
Seems to be a use-case where the next() function is added, when it should not. See screenshot:
Screenshot 2022-04-25 at 13 55 29

Also
- Check the linting error
- add the next() function as optional Typescript type

ErnoW (approved):

PR #336, opened 2022-05-10
Feat/evm api get token allowance

sogunshola:
---
name: GET TOKEN ALLOWANCE
about: A new pull request
---

### Checklist

### Issue Description


packages/evmApi/src/resolvers/token/getTokenAllowance.ts
```diff
@@ -0,0 +1,21 @@
...
+type ApiResult = operations[operation]['responses']['200']['content']['application/json'];
+
+export const getTokenAllowanceResolver = new EvmResolver({
+  getPath: (params: ApiParams) => `erc20/${params.address}/allowance`,
+  apiToResult: (data: ApiResult) => ({
+    ...data
+  }),
+  resultToJson: (data) => ({
+    ...data
+  }),
```
ErnoW, line 19: Ideally we want to return this as an Erc20Value dataType. However for thiswe would also need the 'decimal value of the token.

So for now lets use a Bignumber as internal datatype, and cast it to a string in resultToJson

In the future we might want to accept an Erc20Token as input param. Then we can induce the contractAddress from that dataType to use as params, and use the decimals+symbol in the result.

ErnoW (changes requested):

PR #338, opened 2022-05-10
feat(evmApi): api reslover for endpoint getTokenMetadataBySymbol

sogunshola:
---
name: getTokenMetadataBySymbol
about: A new pull request
---

### Checklist

### Issue Description


packages/evmApi/src/resolvers/token/getTokenMetadataBySymbol.ts
```diff
@@ -0,0 +1,22 @@
...
+
+type QueryParams = operations[operation]['parameters']['query'];
+type ApiParams = QueryParams;
+
+type ApiResult = operations[operation]['responses']['200']['content']['application/json'];
+
+export const getTokenMetadataBySymbolResolver = new EvmResolver({
+  getPath: (params: ApiParams) => `/erc20/metadata/symbols`,
+  apiToResult: (data: ApiResult) =>
+    data.map((token) => ({
```
ErnoW, line 16: Part of the data should be converted to Erc20 datatype (except the things that are not directly related to the Erc20 datatype sunch as "validated")

ErnoW (changes requested):

packages/evmApi/src/utils/getExtraData.ts
```diff
@@ -0,0 +1,10 @@
+export function getExtraData<T, I>(dataType: any, data: I) {
+    let extras = {};
+    (Object.keys(data) as Array<keyof Omit<I, keyof T>>).forEach((key) => {
+      if (dataType[key] == undefined) {
+        extras = { ...extras, [key]: data[key] };
+      }
+    });
+    return extras as Omit<I, keyof T>;
+  }
+
```
ErnoW, line 10: What is the purpose of this function? "getExtraData" is really general, and without any context, I do not get what this does. 

Please consider
- Renaming the function
- Renaming variables (don't use `T`, or `I`)
- don't use any type if possible (try generic types, or resort back to unknown)
- Add comment to describe the purpose of the function and what it serves to do.

packages/evmApi/src/resolvers/token/getTokenMetadataBySymbol.ts
```diff
@@ -0,0 +1,40 @@
...
+  getPath: (params: ApiParams) => `/erc20/metadata/symbols`,
+  apiToResult: (data: ApiResult) =>
+    data.map((token) => {
+      const erc20token = new Erc20Token({
+        ...toCamelCase(token),
+        contractAddress: token.address,
+        // TODO: add chain info (or omit it from Erc20Token type)
+        chain: 1,
+      });
+      const extras = getExtraData<Erc20Input, Camelize<ApiResult[0]>>(erc20token.result, toCamelCase(token));
```
ErnoW, line 29: See my other comment about the function 'getExtraData'

Maybe an alternative is to not use a "getExtraData" function, but rather map each key-value explicitly

ErnoW (changes requested):

PR #352, opened 2022-05-16
Feat/resolve pagination

sogunshola:
---
name: Handle pagination
about: A new pull request
---

### Checklist

### Issue Description


packages/evmApi/src/resolvers/PaginatedResolver.ts
```diff
@@ -0,0 +1,134 @@
...
+import { getNextParams } from '../utils/getNextParams';
+
+type Method = 'get' | 'post';
+
+export interface PaginatedResponse<Data> {
+  total?: number;
+  page?: number;
+  page_size?: number;
+  cursor?: string;
+  result?: Data;
```
ErnoW, line 13: Can these values not be typed as defined values?

packages/evmApi/src/resolvers/PaginatedResolver.ts
```diff
@@ -0,0 +1,134 @@
...
+}
+
+export interface PaginatedOptions {
+  offset?: number;
+  limit?: number;
+  cursor?: string;
```
ErnoW, line 19: Can these values not be typed as defined values?

packages/evmApi/src/EvmApiResultAdapter.ts
```diff
@@ -37,15 +39,27 @@ export type JSONApiResult<Value extends object = object> =
...
   private _adapter: (data: Data) => AdaptedData;
   private _jsonAdapter: (data: AdaptedData) => JSONData;
+  private _nextCall?: (params: Params) => unknown;
 
-  constructor(data: Data, adapter: (data: Data) => AdaptedData, jsonAdapter: (data: AdaptedData) => JSONData) {
+  constructor(
+    data: Data,
+    adapter: (data: Data) => AdaptedData,
+    jsonAdapter: (data: AdaptedData) => JSONData,
+    nextCall?: (params: Params) => unknown,
```
ErnoW, line 57: This needs to be properly typed. Otherwise when someone calls result.next(), they get no typed result back anymore

packages/evmApi/src/EvmApiResultAdapter.ts
```diff
@@ -56,6 +70,21 @@ export class EvmApiResultAdapter<Data extends unknown, AdaptedData extends unkno
     return this._adapter(this._data);
   }
 
+  get next() {
+    return this._nextCall;
+  }
+
+  get pagination(): Partial<Camelize<PaginatedResponse<AdaptedData>>> | null {
```
ErnoW, line 77: I think we can remove the Partial<> modifier, when we move to a specific resultAdapter for paginated responses (as mentioned in previous comment). Then we have better typechecks

packages/evmApi/src/EvmApiResultAdapter.ts
```diff
@@ -56,6 +70,21 @@ export class EvmApiResultAdapter<Data extends unknown, AdaptedData extends unkno
     return this._adapter(this._data);
   }
 
+  get next() {
+    return this._nextCall;
+  }
```
ErnoW, line 75: Lets make this a function that calls _nextCall when defined,
and returns an Error otherwise. This way next() is always typed and availbable for paginated responses

ErnoW: Also `_nextCall` seems to be always defined for paginated results? Where do we have the check that we cannot call next() when the api returns no cursor.

packages/evmApi/src/utils/getNextParams.ts
```diff
@@ -0,0 +1,15 @@
+import { PaginatedOptions, PaginatedResponse } from "../resolvers/PaginatedResolver";
+
+export const getNextParams = <T extends PaginatedOptions, I>(params: T, data: Partial<PaginatedResponse<I>>) => {
```
ErnoW, line 3: This function is difficult to read
- Try not to use one-letter variable names, like `T` and `I`
- Add comments to clarify if things seem not obvious
- If we have strict type checking, then we don't need `if (!data.page_size || !data.total || data.page === undefined) return params;`

ErnoW (changes requested):
Functionality wise looks good. But we need to improve types in some places for better dev experience. 
Also seems logical to split up the resultAdapter in a separate paginated resultAdapter

sogunshola:
Got it.. working on comments now 👍🏾

packages/evmApi/src/resolvers/PaginatedResolver.ts
sogunshola: Regarding this, not all endpoints use cursor right now at the moment.

packages/evmApi/src/resolvers/PaginatedResolver.ts
sogunshola: Also in the generated types, there is no consistency in these options, some endpoints have it defined while most don't

PR #373, opened 2022-05-19
Feat(evm api)/get all token ids

sogunshola:
---
name: getAllTokenIds
about: Fetch and transform token.getAllTokenIds
---

### Checklist

### Issue Description


packages/core/src/dataTypes/tokens/EvmNFTMetadata.ts
```diff
@@ -0,0 +1,60 @@
...
+  name: string;
+  description: string;
+  image: string;
+  animation_url?: string;
+  id?: string;
+  attributes: MoralisDataObjectValue[];
+  createdBy?: string;
+  yearCreated?: string;
+  media: MoralisDataObjectValue;
+}
```
ErnoW, line 13: Where did you get this metadata from?
Every NFT can have its unique metadata, which can be anything

packages/evmApi/src/resolvers/token/getAllTokenIds.ts
```diff
@@ -0,0 +1,43 @@
...
+  getPath: (params: Params) => `nft/${params.address}`,
+  apiToResult: (data: ApiResult) =>
+    data.result?.map((nft) => ({
+      ...toCamelCase(nft),
+      token: new EvmNFT({
+        contractType: nft.contract_type,
+        syncedAt: nft.synced_at,
+        tokenAddress: nft.token_address,
+        tokenId: nft.token_id,
+        tokenUri: nft.token_uri ?? '',
```
ErnoW, line 29: Don't default to an empty string. As an empty string means that is has no tokenUri. So in this case we should accept null as a value in EvmNFT

packages/core/src/dataTypes/tokens/EvmNFT.ts
```diff
@@ -0,0 +1,82 @@
+import { MoralisDataObject } from '../abstract';
+import { EvmAddress, EvmAddressish } from '../EvmAddress';
+import { EvmNFTMetadata } from '.';
```
ErnoW, line 3: import from ./EvmNFTMetadata

packages/evmApi/src/resolvers/token/getAllTokenIds.ts
```diff
@@ -0,0 +1,43 @@
...
+  address: EvmAddressish;
+}
+
+type ApiResult = operations[operation]['responses']['200']['content']['application/json'];
+
+export const getAllTokenIdsResolver = new EvmPaginatedResolver({
+  getPath: (params: Params) => `nft/${params.address}`,
+  apiToResult: (data: ApiResult) =>
+    data.result?.map((nft) => ({
+      ...toCamelCase(nft),
```
ErnoW, line 23: values as syncedAt contractType etc are added twice to the result (one here in the root, and also nested in 'token')

ErnoW (changes requested):

PR #385, opened 2022-05-23
Feat(evm api)/upload folder

sogunshola:
---
name: uploadFolder
about: Fetch and transform storage.uploadFolder
---

### Checklist

### Issue Description


packages/evmApi/src/resolvers/Resolver.ts
```diff
@@ -70,8 +78,13 @@ export class EvmResolver<ApiParams, Params, ApiResult, AdaptedResult, JSONResult
...
         return result;
       }
-      // @ts-ignore TODO: fix the ApiParams type, as it should extend object/record
-      return { ...result, [key]: params[key] };
+      if (this.bodyType === BodyType.PROPERTY) {
+        // @ts-ignore TODO: fix the ApiParams type, as it should extend object/record
+        return { ...result, [key]: params[key] };
+      } else {
+        // @ts-ignore TODO: fix the ApiParams type, as it should extend object/record
+        return params[key];
```
ErnoW, line 86: I don't think this is correct, right?

Why returning only params[key]?

packages/evmApi/src/resolvers/Resolver.ts
```diff
@@ -70,8 +78,13 @@ export class EvmResolver<ApiParams, Params, ApiResult, AdaptedResult, JSONResult
       if (!params[key] || !this.isBodyParam(key)) {
         return result;
       }
-      // @ts-ignore TODO: fix the ApiParams type, as it should extend object/record
-      return { ...result, [key]: params[key] };
+      if (this.bodyType === BodyType.PROPERTY) {
+        // @ts-ignore TODO: fix the ApiParams type, as it should extend object/record
+        return { ...result, [key]: params[key] };
+      } else {
```
ErnoW, line 84: These if-else checks can become cleaner.

Can you enable the `no-else-return` eslint rule and change it accordingly? 

Reference: https://eslint.org/docs/rules/no-else-return

ErnoW (changes requested):

packages/evmApi/src/resolvers/Resolver.ts
sogunshola: `params[key]` is returned because when the body of the post request is not a known object (or maybe an array), the body is collected from the parameters with the key `data` and the content of `data` is used as the body.

PR #392, opened 2022-05-25
Feat(evm api)/server requestand evm sync

sogunshola:
---
name: PR handling 3 issues
about: Issues 274 | 273| 248
---

### Checklist

### Issue Description


demos/test-app-all/src/EvmApi.tsx
```diff
@@ -5,6 +5,14 @@ export const EvmApi = () => {
...
       <h2>EVM Api</h2>
 
+      <button
+        onClick={() => {
+          Moralis.EvmApi.start();
+        }}
+      >
+        start
+      </button>
+
```
ErnoW, line 15: Not needed, as we already call Moralis.start in App.tsx

demos/test-app-all/src/EvmApi.tsx
```diff
@@ -70,7 +78,7 @@ export const EvmApi = () => {
         onClick={() => {
           Moralis.EvmApi.native
             .getLogsByAddress({
-              address: '0xa2107fa5b38d9bbd2c461d6edf11b11a50f6b974',
+              // address: '0xa2107fa5b38d9bbd2c461d6edf11b11a50f6b974',
             })
             .then(console.log);
```
ErnoW, line 83: It does not make sense to prefill the user address here.
As the address in this request is refering to the smart contract address
See: https://docs.moralis.io/moralis-dapp/web3-api/native?q=runContractFunction

Can you only prefill the address in requests where it is refering to to the useraddress?

packages/evmApi/src/resolvers/Resolver.ts
```diff
@@ -1,8 +1,16 @@
-import core, { RequestController } from '@moralisweb3/core';
+import core, { ApiErrorCode, EvmAddress, EvmChain, MoralisApiError, RequestController } from '@moralisweb3/core';
 import { BASE_URL } from '../EvmApi';
 import { EvmApiResultAdapter } from '../EvmApiResultAdapter';
 
 type Method = 'get' | 'post';
+export enum BodyType {
+  PROPERTY = 'property',
+  BODY = 'set body',
+}
```
ErnoW, line 9: I'm missing some context here. What and why is this used for?

packages/evmApi/src/resolvers/Resolver.ts
```diff
@@ -114,7 +139,56 @@ export class EvmResolver<ApiParams, Params, ApiResult, AdaptedResult, JSONResult
...
+        code: ApiErrorCode.GENERIC_API_ERROR,
+        message: 'EvmApi failed: start with apiKey or serverUrl',
+      });
+    }
+    return `${serverUrl}/functions/${this.name}`;
+  }
+
+  protected resolveDefaultParams(params: ApiParams) {
+    const evm = core.modules.getNetwork('evm');
+    // @ts-ignore TODO: fix type as it should be of type MoralisEvm
```
ErnoW, line 173: These errors can be fixed by extending the type of NetworkModule. If you add `isConnected` as a method in there

packages/evmApi/src/resolvers/Resolver.ts
```diff
@@ -114,7 +139,56 @@ export class EvmResolver<ApiParams, Params, ApiResult, AdaptedResult, JSONResult
...
+    }
+    // @ts-ignore
+    if (evm && evm.isConnected) {
+      // @ts-ignore TODO: fix types for params (probably extend a default options interface)
+      params.chain = params.chain ?? EvmChain.create(evm.chain).apiHex;
+      // @ts-ignore
+      params.address = params.address ?? EvmAddress.create(evm.account).lowercase;
+    }
+    return params;
+  }
```
ErnoW, line 188: We might actually want to resolve default values on implementation, instead of having it generically on ALL requests.

As I mentioned in an earlier comment; for some requests it makes no sense to pre-fill the address with the userAddress (because the address refers to a contract address).

ErnoW: So possibly we could do something like:
```js
export const getTokenPriceResolver = new EvmResolver({
  ...,
  parseParams: (params: Params) => ({
    // -> Here
    chain: resolveDefaultChain(params.chain),
    address: resolveDefaultAddress(params.address),

    exchange: params.exchange,
    to_block: params.toBlock,
    providerUrl: params.providerUrl,
  }),
});

```

That gives us a lot of flexibility, on where and how we want to resolve the address.

demos/test-app-all/src/EvmApi.tsx
sogunshola: Disregard, was using to run some tests locally. removing in next push

packages/evmApi/src/resolvers/Resolver.ts
sogunshola: Taking this out. belongs in a separate MR

packages/evmApi/src/utils/resolveDefaultParams.ts
```diff
@@ -0,0 +1,30 @@
+import core, {
+  ApiErrorCode,
+  EvmAddress,
+  EvmAddressish,
+  EvmChain,
+  EvmChainish,
+  MoralisApiError,
+} from '@moralisweb3/core';
+
+export const resolveDefaultAddress = (address?: EvmAddressish) => {
```
ErnoW, line 10: It makes more sense to let this function return an EvmAddress instead of a string.
- Better usable in the future
- Corresponds better with the name of this function. We handle any address as an EvmAddress. Unless we are parsing to an output.

packages/evmApi/src/utils/resolveDefaultParams.ts
```diff
@@ -0,0 +1,30 @@
...
+  const evm = core.modules.getNetwork('evm');
```
ErnoW, line 11: This throws an error if there is no `evm` package registered.

packages/evmApi/src/utils/resolveDefaultParams.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (address) {
+    return EvmAddress.create(address).lowercase;
+  } else if (!evm || !evm.account) {
+    throw new MoralisApiError({
+      code: ApiErrorCode.GENERIC_API_ERROR,
+      message: 'EvmApi failed: address is required',
+    });
+  } else {
+    return EvmAddress.create(evm.account).lowercase;
```
ErnoW, line 20: Please enable the `no-else-return` eslint and clean this up: https://eslint.org/docs/rules/no-else-return

We should write better if/else checks that are more readable and maintainable that way.

packages/evmApi/src/utils/resolveDefaultParams.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+};
+
+export const resolveDefaultChain = (chain?: EvmChainish) => {
+  const evm = core.modules.getNetwork('evm');
```
ErnoW, line 25: This can throw an error if no evm package is installed

packages/evmApi/src/utils/resolveDefaultParams.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (evm && evm.chain && !chain) {
+    return EvmChain.create(evm.chain).apiHex;
+  }
+  return chain ? EvmChain.create(chain).apiHex : undefined;
```
ErnoW, line 29: Lets resolve to eth as default chain

packages/evmApi/src/resolvers/storage/uploadFolder.ts
```diff
@@ -0,0 +1,28 @@
+import { operations } from '../../generated/types';
```
ErnoW, line 1: Should not be part of this PR, as it is submitted in another PR already.

packages/evmApi/src/resolvers/storage/index.ts
```diff
@@ -0,0 +1 @@
+export { uploadFolderResolver } from './uploadFolder';
```
ErnoW, line 1: Should not be part of this PR, as it is submitted in another PR already.

packages/evmApi/src/resolvers/native/getContractEvents.ts
```diff
@@ -0,0 +1,46 @@
+import { EvmChainish, EvmAddress, EvmAddressish } from '@moralisweb3/core';
```
ErnoW, line 1: Should not be part of this PR

ErnoW (changes requested):

PR #414, opened 2022-05-30
feat: improve Transaction datatype and

ErnoW:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description
- Split up and cleanup files based on EvmTransactions
- Add docs for evm transaction properties
- Improve .equals functions for evm transactions and make more consistent
- Fix todo where we pass chain: 1, for all transaction responses in the api. Now we use the 'chain' param to supply the correct chain to the evnTransactions

### DIFF
--- a/packages/evmApi/src/EvmApiPaginatedResultAdapter.ts
+++ b/packages/evmApi/src/EvmApiPaginatedResultAdapter.ts
@@ -12,7 +12,7 @@ export class EvmApiPaginatedResultAdapter<
 
   constructor(
     data: Data,
-    adapter: (data: Data) => AdaptedData,
+    adapter: (data: Data, params: Params) => AdaptedData,
     jsonAdapter: (data: AdaptedData) => JSONData,
     params: Params,
     nextCall?: () => Promise<EvmApiPaginatedResultAdapter<Data, AdaptedData, JSONData, Params>>,
--- a/packages/evmApi/src/EvmApiPaginatedResultAdapter.ts
+++ b/packages/evmApi/src/EvmApiPaginatedResultAdapter.ts
@@ -6,16 +6,18 @@ export class EvmApiPaginatedResultAdapter<
   Data extends PaginatedResponse<unknown>,
   AdaptedData,
   JSONData,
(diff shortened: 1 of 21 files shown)


packages/evmApi/src/resolvers/account/getTokenBalances.ts
```diff
@@ -21,11 +26,15 @@ export const getTokenBalancesResolver = new EvmResolver({
         contractAddress: token.token_address,
         logo: token.logo,
         thumbnail: token.thumbnail,
-        // TODO: add chain info (or omit it from Erc20Token type)
-        chain: 0,
+        // TODO: Fix typing that chain always is set (because we have default value in parseParams)
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        chain: params.chain!,
```
sogunshola, line 31: Great idea getting the chain back from the params 🔥

packages/evmApi/src/EvmApiPaginatedResultAdapter.ts
```diff
@@ -6,16 +6,18 @@ export class EvmApiPaginatedResultAdapter<
...
   JSONData,
-> extends EvmApiResultAdapter<Data, AdaptedData, JSONData> {
-  private _nextCall?: () => Promise<EvmApiPaginatedResultAdapter<Data, AdaptedData, JSONData>>;
+  Params,
+> extends EvmApiResultAdapter<Data, AdaptedData, JSONData, Params> {
+  private _nextCall?: () => Promise<EvmApiPaginatedResultAdapter<Data, AdaptedData, JSONData, Params>>;
 
   constructor(
     data: Data,
     adapter: (data: Data) => AdaptedData,
```
sogunshola, line 15: I think the `adapter` here should also accept `Params` like in `EvmApiResultAdapter.ts`

sogunshola:
```suggestion
    adapter: (data: Data, params: Params) => AdaptedData,
```

PR #440, opened 2022-06-02
code clean up, remove few todos and little optimisations

sogunshola:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description


packages/evmApi/src/resolvers/PaginatedResolver.ts
```diff
@@ -92,14 +92,12 @@ export class EvmPaginatedResolver<
...
-    const bodyParams = this.getBodyParams(apiParams);
-
     const { result } = await RequestController.post<
       ServerResponse<PaginatedResponse<ApiResult>>,
       Record<string, string>,
-      Record<string, string>
-    >(url, searchParams, bodyParams);
+      //@ts-ignore TODO: fix the ApiParams type, as it should extend object/record
+      ApiParams
+    >(url, {}, apiParams);
```
ErnoW, line 100: Why did you remove the searchParams? 
We can have post requests that have both search params and bodyParams.

packages/evmApi/src/resolvers/Resolver.ts
```diff
@@ -159,14 +159,12 @@ export class EvmResolver<ApiParams, Params, ApiResult, AdaptedResult, JSONResult
...
 
-    const searchParams = this.getSearchParams(apiParams);
-    const bodyParams = this.getBodyParams(apiParams);
-
-    const { result } = await RequestController.post<
-      ServerResponse<ApiResult>,
-      Record<string, string>,
-      Record<string, string>
-    >(url, searchParams, bodyParams);
+    //@ts-ignore TODO: fix the ApiParams type, as it should extend object/record
```
ErnoW, line 162: Why did you remove the searchParams?
We can have post requests that have both search params and bodyParams.

packages/evmApi/src/resolvers/defi/getPairAddress.ts
```diff
@@ -22,7 +22,7 @@ export const getPairAddressResolver = new EvmResolver({
   name: 'getPairAddress',
   getPath: (params: Params) => `${params.token0Address}/${params.token1Address}/reserves`,
   apiToResult: (data: ApiResult, params: Params) => ({
-    //   ApiResult types generated all come as undefined which should not be the case TODO:
+    //   ApiResult types generated all come as undefined which should not be the case
```
ErnoW, line 25: Please leave this TODO as it is something that is not addressed yet.

packages/evmApi/src/resolvers/token/getAllTokenIds.ts
```diff
@@ -15,11 +15,30 @@ export interface Params extends Camelize<Omit<ApiParams, 'chain' | 'address'>>,
 }
 
 type ApiResult = operations[operation]['responses']['200']['content']['application/json'];
+interface Result extends Omit<ApiResult, 'result'> {
```
ErnoW, line 18: Can you add a TODO to revisit this when api has updated the definition. Also add an issue ticket for this, so we can pick this up next sprint.

Also do not name this Result but rather use ApiResult as it might cause confusion that this is a parsed result.

You can name the original `ApiResult` to `GeneratedApiResult`

packages/evmApi/src/resolvers/token/getTokenIdMetadata.ts
```diff
@@ -17,10 +17,17 @@ export interface Params extends Camelize<Omit<ApiParams, 'chain' | 'address'>> {
 
 type ApiResult = operations[operation]['responses']['200']['content']['application/json'];
 
+export interface Result extends ApiResult {
```
ErnoW, line 20: Can you add a TODO to revisit this when api has updated the definition. Also add an issue ticket for this, so we can pick this up next sprint.

Also do not name this Result but rather use ApiResult as it might cause confusion that this is a parsed result.

You can name the original `ApiResult` to `GeneratedApiResult`

packages/server/src/utils/parseFile.ts
```diff
@@ -0,0 +1,20 @@
+import { File, FullOptions } from 'parse';
```
ErnoW, line 1: Why is this file included?

ErnoW (changes requested):

packages/evmApi/src/resolvers/Resolver.ts
sogunshola: I only made this update for server requests. The server requests are all POST and the parameters are all expected in the body.

packages/evmApi/src/resolvers/PaginatedResolver.ts
sogunshola: I only made this update for server requests. The server requests are all POST and the parameters are all expected in the body.

packages/server/src/utils/parseFile.ts
sogunshola: taking out, was not meant to be part of the PR

PR #453, opened 2022-06-14
feat(evm): validate optional chain

sogunshola:
When calling transferNative, and any other Evm functionalities, we can
let the user provide a chain (EvmChainish). When they do, we check if
the provided chain corresponds the current chain. If that is not the
case we throw an error.

---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

### DIFF
--- a/packages/evm/src/chainMethods/transferErc721.ts
+++ b/packages/evm/src/chainMethods/transferErc721.ts
@@ -1,8 +1,10 @@
 import { EvmAddress, EvmAddressish, EvmChainish } from '@moralisweb3/core';
 import { BigNumberish, ethers } from 'ethers';
 import { assertAddress } from '../assert/assertAddress';
+import { assertChain } from '../assert/assertChain';
 import { assertProvider } from '../assert/assertProvider';
 import { Erc721__factory } from '../Contract';
+import { validateChain } from '../utils/validators';
 import { wrapEthersTransactionResponse } from '../utils/wrapEthersTransactionResponse';
 
 export interface TransferErc721Options {
@@ -18,7 +20,8 @@ export const makeTransferErc721 =
     const provider = assertProvider(_provider);
     const fromAddress = assertAddress(_account, 'No account is connected');
 
-    // TODO: validate that provided chain is current chain
+    const chain = assertChain(_chain, 'Chain is not set on MoralisEvm. Make sure to be properly connected');
+    if (options.chain) validateChain(options.chain, chain);
 
     const contractAddress = EvmAddress.create(options.contractAddress);
     const toAddress = EvmAddress.create(options.to);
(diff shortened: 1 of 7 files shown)


packages/evm/src/utils/validators.ts
```diff
@@ -0,0 +1,10 @@
+import { EvmChain, EvmChainish, MoralisNetworkError, NetworkErrorCode } from '@moralisweb3/core';
+
+export const validateChain = (providedChain: EvmChainish, chain: EvmChain): void => {
+  if (!chain.equals(providedChain)) {
+    throw new MoralisNetworkError({
+      code: NetworkErrorCode.CHAIN_MISMATCH,
+      message: 'Provided chain does not match evm connected chain',
```
ErnoW, line 7: This is not a generic message, so this  function is not a reusable util.
Make it so we can provide a message, and default to a generic message like "Expected chain "xxx", but got "yyy"). You probably can use the chain.hex values for this

packages/evm/src/utils/validators.ts
```diff
@@ -0,0 +1,10 @@
+import { EvmChain, EvmChainish, MoralisNetworkError, NetworkErrorCode } from '@moralisweb3/core';
+
+export const validateChain = (providedChain: EvmChainish, chain: EvmChain): void => {
```
ErnoW, line 3: The purpose of this function is pretty similar to assertChain. Consider moving this function there for easier discovery.

Also naming can be better, consider renaming it to assertChainEquality (I was confused at first whay we needed this function, as we already have assertChain)

packages/evm/src/chainMethods/transferNative.ts
```diff
@@ -21,8 +22,8 @@ export const makeTransferNative =
   async (options: TransferNativeOptions) => {
     const to = EvmAddress.create(options.to);
     const value = EvmNative.create(options.value);
-    // TODO: check if provided chain is same as current chain, otherwise throw error
     const chain = assertChain(_chain, 'Chain is not set on MoralisEvm. Make sure to be properly connected');
+    if (options.chain) validateChain(options.chain, chain);
```
ErnoW, line 26: Can you add curly braces? that's easier to read.

I will add an eslint rule for this in a separate PR

```suggestion
    if (options.chain) {
      validateChain(options.chain, chain);
    }
```

ErnoW: Same for the other files

packages/evm/src/chainMethods/transferErc721.ts
```diff
@@ -18,7 +20,8 @@ export const makeTransferErc721 =
     const provider = assertProvider(_provider);
     const fromAddress = assertAddress(_account, 'No account is connected');
 
-    // TODO: validate that provided chain is current chain
+    const chain = assertChain(_chain, 'Chain is not set on MoralisEvm. Make sure to be properly connected');
```
ErnoW, line 23: 👍

ErnoW (changes requested):

ErnoW (dismissed):

ErnoW (approved):

PR #459, opened 2022-06-19
Feat/auth web3auth

sogunshola:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

### DIFF
--- a/packages/evmConnectors/EvmWeb3authConnector/package.json
+++ b/packages/evmConnectors/EvmWeb3authConnector/package.json
@@ -1,0 +1,30 @@
+{
+  "name": "@moralisweb3/evm-web3auth-connector",
+  "author": "Moralis",
+  "version": "2.0.0-alpha.1",
+  "license": "MIT",
+  "private": false,
+  "main": "./lib/index.js",
+  "types": "./lib/index.d.ts",
+  "sideEffects": false,
+  "scripts": {
+    "format": "prettier . '**/*.+(js|ts|json)' --write",
+    "format:check": "prettier . '**/*.+(js|ts|json)' --check",
+    "lint": "eslint . --ext .js,.ts,.tsx,jsx",
+    "clean": "rm -rf lib && rm -rf lib.esm && rm -rf tsconfig.tsbuildinfo",
+    "build": "tsc",
+    "dev": "tsc --watch"
+  },
+  "devDependencies": {
+    "@walletconnect/types": "^1.7.5",
+    "eslint": "^8.7.0",
+    "prettier": "^2.5.1",
+    "typescript": "^4.5.5"
+  },
+  "dependencies": {
+    "@moralisweb3/core": "^2.0.0-alpha.1",
+    "@moralisweb3/evm-connector-utils": "^2.0.0-alpha.1",
+    "ethers": "^5.6.2",
+    "@web3auth/web3auth": "0.5.1"
+  }
+}
--- a/packages/evmConnectors/EvmWeb3authConnector/src/EvmWeb3authConnector.ts
+++ b/packages/evmConnectors/EvmWeb3authConnector/src/EvmWeb3authConnector.ts
@@ -1,0 +1,93 @@
+import {
+  EvmAddress,
+  EvmChain,
+  EvmConnectResponse,
+  MoralisNetworkConnectorError,
+  EvmWeb3authConnectOptions,
+  NetworkConnectorErrorCode,
+} from '@moralisweb3/core';
+import core from '@moralisweb3/core';
+import { EvmAbstractConnector } from '@moralisweb3/evm-connector-utils';
+import { ethers } from 'ethers';
+import { Web3Auth, Web3AuthOptions } from '@web3auth/web3auth';
+import { SafeEventEmitterProvider } from '@web3auth/base';
+
+/**
+ * Connector for WalletConnect v1
+ */
+export class EvmWeb3authConnector extends EvmAbstractConnector {
+  constructor() {
+    super({
+      name: 'web3auth',
+      core,
+    });
+  }
+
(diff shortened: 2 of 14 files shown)


packages/evmConnectors/EvmWeb3authConnector/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "devDependencies": {
+    "@walletconnect/types": "^1.7.5",
+    "eslint": "^8.7.0",
+    "prettier": "^2.5.1",
+    "typescript": "^4.5.5"
+  },
+  "dependencies": {
+    "@moralisweb3/core": "^2.0.0-alpha.1",
+    "@moralisweb3/evm-connector-utils": "^2.0.0-alpha.1",
+    "ethers": "^5.6.2",
```
ErnoW, line 27: I don't think we need this dependency. Lets avoid this.

packages/evmConnectors/EvmWeb3authConnector/src/EvmWeb3authConnector.ts
```diff
@@ -0,0 +1,93 @@
...
+        message: 'Failed to create a WalletConnect provider',
+      });
+    }
+
+    return provider;
+  }
+
+  async connect(_options: EvmWeb3authConnectOptions): Promise<EvmConnectResponse> {
+    const chainId = EvmChain.create(_options.chainId || '0x1').apiHex;
+    const options: Web3AuthOptions = {
```
ErnoW, line 50: Lets define the default values seperately (look at EvmMetamaskConnector for example). That's better readable. then we can merge the provided options and default options here

packages/evmConnectors/EvmWeb3authConnector/src/EvmWeb3authConnector.ts
```diff
@@ -0,0 +1,93 @@
...
+    this.logger.verbose('Connecting', { providedOptions: _options, options });
+
+    const web3auth = new Web3Auth(options as Web3AuthOptions);
+
+    await web3auth.initModal();
+
+    await this.getProvider(web3auth);
+
+    const isSocialLogin = web3auth.provider ? false : true;
+    const ether = new ethers.providers.Web3Provider(web3auth.provider || (web3auth as any));
```
ErnoW, line 73: What's the benefit of parsing it already to an ethers Web3Provider here? With other connectors we do that only in the Evm Module.

I think web3Auth already returns an EIP1193 provider, which is all we care about.

This also prevents us to import a big dependency as ethers for this connector

sogunshola: This is mainly because the web3Auth provider is only available if a user uses a wallet to authenticate. in a case where they use google or any social platform, provider is undefined.

ErnoW: If the provider is undefined, then how does it work when we create a provider with `ethers.providers.Web3Provider`? This looks to me like it would fail as well.

Is it intended behavior of web3Auth to return undefined in these cases? Or is there an error in our/their flow?


Any connection to Evm should be based on some sort of provider that is responsible for the connection, most likely this is EIP1193 (but could be an older format in some cases). That's the provider that we need, and is used in the Evm package after a successful connection.

sogunshola: The behaviour I noticed is when I connect using google, the web3auth.provider is `null`. 
`await web3auth.connect()` returns a provider or null, null in a case of where socials is used and returns provider if a wallet is used.
Looking at what was implemented in sdkv1 we did
`const ether = new ethers.providers.Web3Provider(
        web3auth?.provider ? web3auth.provider : web3auth
      );` which surprisingly works.
      
I went through web3auth docs and they used web3 not ethers, which they used to get address 
https://web3auth.io/docs/guides/web3auth-eth

ErnoW: > surprisingly works

We need to know why it works. That's what I'm aiming at. So far we are initialising `ethers.providers.Web3Provider` via EIP1193 providers (returned by Metamask, Walletconnect and also Web3Auth when using non-social methods, right?).

But if `web3auth?.provider` is not set, then we are initialising an ethers provider by using `webauth` as input? That part is confusing to me. Is that object also EIP1193? Or is it an web3-compatible provider that we can pass into ethers.providers.Web3Provider (https://docs.ethers.io/v5/api/providers/other/#Web3Provider)?

- If `web3auth` is already EIP1193 compatible, then we can just return that provider (without wrapping it in `new ethers.providers.Web3Provider` (this will be done in Evm module)
- If `web3auth` is another type of web3-compatible provider, then we need to see if we can convert it to EIP1193, or we could support multiple standards as value for for the `connector.provider`

sogunshola: You’re right, if it initializes ethers, then it’s some kind of provider. I’ll figure out why it behaves that way.

sogunshola: So I figured that the issue is `provider?.request({ method: 'eth_requestAccounts' })` . this request fails because there is no account when you use social login but works when you use a wallet as that already has an account.

I also figured out that web3auth.connect()  returns an eip1193 provider regardless of the method used to login which makes `ethers.provider,web3provider` work perfectly.

Right now we need a way to create a wallet without using ethersjs or web3js.

sogunshola: Seems like we might have to resolve to using either web3js or ethersjs for this.

PR #469, opened 2022-07-01
feat(cdn builds): usage in browsers via a script tag

b4rtaz:
---
name: CDN builds
about: Usage in browsers via a script tag
---

### Checklist


packages/core/package.json
```diff
@@ -7,6 +7,10 @@
   "main": "./lib/index.js",
   "types": "./lib/index.d.ts",
   "sideEffects": false,
+  "files": [
+    "lib",
+    "dist"
+  ],
```
ErnoW, line 13: Should we also specify the main and module exports, here in package.json?
For example as is described in this discussion: https://stackoverflow.com/questions/72457791/typescript-packages-that-ship-with-mjs-and-d-ts-but-without-d-mts-how-to-i] (first answer).

Not sure if it is required, but this definition seems common when packages ship CJS and ESM

packages/evmConnectors/walletConnectWrapper/README.md
```diff
@@ -0,0 +1,5 @@
+# @moralisweb3/wallet-connect-wrapper
+
+Rollup cannot build an UMD bundle with `@walletconnect/web3-provider`, so this project is a workaround for this problem. It contains a wrapper of WalletConnectProvider. How it works? Webpack creates a single CommonJs bundle which Rollup can consume without any problem.
+
+The idea is taken from [walletconnect-v1](https://github.com/DePayFi/walletconnect-v1).
```
ErnoW, line 5: Too bad that they offer no good UMD support, this seems like an acceptable solution to this. 👍

packages/moralis/src/index.ts
```diff
@@ -2,16 +2,26 @@
...
-const Moralis = { Core, Server, Evm, EvmApi, start };
+const Moralis = {
+  Core,
+  Server,
+  Evm,
+  EvmApi,
+  start,
+
+  // for UMD bundles
+  CoreLib,
```
ErnoW, line 23: Why is this only needed for the our core package? Is it because we are exporting named and default exports? We could change it to only export named exports from any our our packages.

ErnoW (changes requested):
Nice work. I have left a few questions.

Also main issue currently is that the CDN test app does not work.
I run `yarn` and `yarn build` following by `yarn e2e`(in test-cdn). But I get the error:
```
"GET /packages/moralis/dist/index.umd.js" Error (404): "Not found"
```
Which seems correct, as no `index.umd.js` file has been created during `yarn build`

.vscode/settings.json
```diff
@@ -3,5 +3,6 @@
   "search.exclude": {
     "**/.git": true,
     "**/node_modules": true
-  }
+  },
+  "liveServer.settings.port": 5502
```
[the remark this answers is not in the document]
b4rtaz: I agree.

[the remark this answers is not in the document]
b4rtaz: I agree.

demos/test-app-all/package.json
```diff
@@ -9,6 +9,8 @@
     "craco": "^0.0.3",
     "jest": "26.6.0",
     "moralis": "^2.0.0-alpha.1",
+    "@moralisweb3/core": "^2.0.0-alpha.1",
+    "@moralisweb3/server": "^2.0.0-alpha.1",
```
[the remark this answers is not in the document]
b4rtaz: I agree.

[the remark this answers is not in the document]
b4rtaz: I agree.

packages/moralis/src/index.ts
b4rtaz: We can don't place `core` + `evm-connector-utils` into UMD builds by that trick. These packages are imported from `Moralis.CoreLib` and `Moralis.EvmConnectorUtilsLib`. An another advantage a smaller bundle size.

Now we have only one case here `packages/evmConnectors/EvmWalletconnectConnector/rollup.config.js`.

```js
...
const umdExternalMap = {
  '@moralisweb3/core': 'Moralis.CoreLib',
  '@moralisweb3/evm-connector-utils': 'Moralis.EvmConnectorUtilsLib',
};
...
```

In the `dist/index.umd.js` build now we have:

<pre>
(function (global, factory) {
    ...
    (global = (...) = <strong>factory(global.Moralis.CoreLib, global.Moralis.EvmConnectorUtilsLib)</strong> );
})(this, (function (core, evmConnectorUtils){
...
</pre>

packages/core/package.json
b4rtaz: As we discussed, it's too early for ESM packages.

ErnoW (approved):

PR #470, opened 2022-07-01
Add mockconnector for tests

ErnoW:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

- Testing Evm connection by using a MockConnector
- Update test scripts for better debugging


packages/integration/src/MockConnector.ts
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  setTx(tx: Partial<Tx>) {
+    this.tx = { ...this.tx, ...tx };
+  }
+
+  /**
+   * Resolve all rpc requests
+   */
+  request(args: RequestArguments) {
```
b4rtaz, line 149: Maybe we should consider a single policy for a method notation. I've seen in the project different policies. So it's hard for newbies to choose correct one.

My suggestion is to enable `explicit-function-return-type` + `explicit-member-accessibility` for EsLint.

packages/integration/package.json
```diff
@@ -9,9 +9,9 @@
   "types": "./lib/index.d.ts",
   "sideEffects": false,
   "scripts": {
-    "test": "cross-env NODE_OPTIONS=--experimental-vm-modules yarn jest --runInBand --detectOpenHandles --forceExit",
+    "test": "cross-env NODE_OPTIONS=--experimental-vm-modules yarn jest --runInBand --detectOpenHandles --forceExit --no-cache",
     "test:ci": "yarn test --ci",
-    "test:old": "yarn jest",
+    "test:debug": "node --trace-warnings --experimental-vm-modules ../../node_modules/jest/bin/jest.js --runInBand --detectOpenHandles --forceExit --no-cache",
```
b4rtaz, line 14: It's still don't fix the problem. Check the case below.

[1] Add to `packages/integration/test/core/config/readConfig.test.ts` this line: `var q = 1;` to test method (anywhere). It's a minor mistake (unused variable).
[2] Run `yarn run test`.
[3] Jest is stuck on this test.

I described [here](https://github.com/MoralisWeb3/Moralis-JS-SDK/tree/test/evm-metamaskConnection/packages/integration#readme) the way how to find an error. But it's time consuming.

sogunshola (dismissed):
I support Bart's suggestion about considering having a pattern for methods, however it might require changes in many files (as enabling rules will throw warnings in so many places) since a lot has been done already, so if we are to do it, it's best we decide asap.

Asides that, code looks good 👍🏾

PR #471, opened 2022-07-05
feat(custom connectors): documentation for custom connectors + simplify base class.

b4rtaz:
---
name: Custom Connectors
about: Documentation for custom connectors + simplify base class.
---

### Checklist


ErnoW (approved):
Looks good, the only concerns are:
- We need good tests for this Abstract connector
- We need to make the NetworkModule independent of Evm logic

But we can pick up these tasks in another PR.

PR #480, opened 2022-07-08
test(core): evmTransaction

sogunshola:
test coverate for evmTransaction
---
name: evmTransaction Test
about: A new pull request
---

### Checklist

### Issue Description


packages/core/src/test/EvmTransaction.test.ts
```diff
@@ -0,0 +1,135 @@
+// Set variables
```
b4rtaz, line 1: An unnecessary comment.

packages/core/src/test/EvmTransaction.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  it('should generate eth request object with only required fields', () => {
+    const transaction = EvmTransaction.create({ chain: 1 });
+    const request = transaction.toEthRequest();
+
+    expect(request.chainId).toBe(1);
+    expect(request.value).toBeUndefined();
+  });
+
+  it('should call the send function', async () => {
```
b4rtaz, line 110: This test doesn't proof that, the callback method was called. You may do it in this way:

```
const x = jest.fn(() => { return '123'; });

x();

expect(x).toBeCalledTimes(1);
```

packages/core/src/test/EvmTransaction.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    expect(() => transaction.send()).toThrowErrorMatchingInlineSnapshot(
+      `"[C0001] Cannot send transaction, no supported call method provided"`,
+    );
+  });
+
+  it('should format EvmTransaction', () => {
+    const transaction = EvmTransaction.create(inputWithAllData);
+    const request = transaction.format();
+
+    expect(request.to).toBe(inputWithAllData.to);
```
b4rtaz, line 133: Too few expectations.

b4rtaz (changes requested):

packages/core/src/test/EvmTransaction.test.ts
```diff
@@ -111,11 +109,11 @@ describe('EvmTransaction', () => {
...
     );
+    const response = jest.fn(async () => await transactionWithSendCall.send());
 
-    const response = await transactionWithSendCall.send();
+    response();
 
-    expect(response.result.hash).toBe('0x123');
-    expect(response.result.chain?.decimal).toBe(inputWithAllData.chain);
+    expect(response).toBeCalledTimes(1);
   });
```
b4rtaz, line 117: More something like this:

```js

const callSpy = jest.fn(async (transaction) =>
    EvmTransactionResponse.create({ ...inputWithAllData, hash: '0x123', from: transaction.result.from }),
);
const transactionWithSendCall = EvmTransaction.create(inputWithAllData, callSpy);

const response = await transactionWithSendCall.send();

expect(response.result.hash).toBe('0x123');
expect(response.result.chain?.decimal).toBe(inputWithAllData.chain);
expect(callSpy).toBeCalledTimes(1);
```

In this way, you check that, the callSpy method is called exactly one time.

b4rtaz (changes requested):

packages/core/src/test/EvmTransaction.test.ts
sogunshola: Ahh. got it.  Thank you

b4rtaz (approved):

PR #489, opened 2022-07-12
test(core): test coverage for EvmTransactionResponse

sogunshola:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description


packages/core/src/test/EvmTransactionResponse.test.ts
```diff
@@ -0,0 +1,116 @@
+// Set variables
```
b4rtaz, line 1: Should this comment be below the import section?

packages/core/src/test/EvmTransactionResponse.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    const equality = transactionResponseA.equals(transactionResponseB);
+
+    expect(equality).toBe(false);
+  });
+
+  it('should convert EvmTransactionResponse to JSON', () => {
+    const transactionResponse = EvmTransactionResponse.create(inputWithAllData);
+    const result = transactionResponse.toJSON();
+
+    expect(result.to).toBe(inputWithAllData.to);
```
b4rtaz, line 108: in this place, I think we need more expectations.

packages/core/src/test/EvmTransactionResponse.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    expect(equality).toBe(true);
+  });
+
+  it('should confirm inequality of 2 EvmTransactionResponses', () => {
+    const transactionResponseA = EvmTransactionResponse.create(inputWithAllData);
+    const transactionResponseB = EvmTransactionResponse.create(requiredParamsOnly);
+    const equality = transactionResponseA.equals(transactionResponseB);
+
+    expect(equality).toBe(false);
+  });
```
ErnoW, line 102: Can you add more cases for equality checks?
If you run `yarn test:coverage` then you can see line coverage. Some cases are not yet included in the unit tests

ErnoW (changes requested):
Nice work, please add more coverage for the equality tests.
Also the wait function is not covered.

run `yarn test:coverage` in the core package to get a coverage report.

packages/evmUtils/src/dataTypes/EvmTransactionResponse/EvmTransactionResponse.test.ts
```diff
@@ -0,0 +1,196 @@
...
+  gasPrice: 21000,
+  gasUsed: 21000,
+  logs: [],
+  status: 1,
+};
+
+describe('EvmTransactionResponse', () => {
+  beforeAll(() => {
+    setupEvmUtils();
+  });
```
b4rtaz, line 54: It this necessary?

sogunshola: Yes. the module has to be registered first. That's the case in all tests for evmUtils

b4rtaz: But you don't test the module here. You test just only one class without a reference to modules. If the module setup is not necessary to test the class, you shouldn't call it. Tests should touch only minimal region of the code.

b4rtaz (approved):

packages/evmUtils/src/dataTypes/EvmTransactionResponse/EvmTransactionResponse.test.ts
b4rtaz: After @sogunshola feedback: this setup is necessary.

b4rtaz (approved):

PR #492, opened 2022-07-13
test(evm): testing transferErc20

Golden-C:
---
name: Integration test for transacting Erc20
about: Testing making transfer Erc20 transaction
---

### Checklist

### Issue Description

### DIFF
--- a/packages/integration/test/evm/transferErc20.test.ts
+++ b/packages/integration/test/evm/transferErc20.test.ts
@@ -1,0 +1,29 @@
+import Core, { Erc20Value, EvmNative } from '@moralisweb3/core';
+import Evm from '@moralisweb3/evm';
+import MockEvmConnector from '../../src/MockConnector';
+
+describe('Evm connect', () => {
+  beforeAll(async () => {
+    Core.registerModules([Evm]);
+    Evm.connectors.register(MockEvmConnector);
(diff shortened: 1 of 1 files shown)


packages/integration/test/evm/transferErc20.test.ts
```diff
@@ -0,0 +1,29 @@
+import Core, { Erc20Value, EvmNative } from '@moralisweb3/core';
+import Evm from '@moralisweb3/evm';
+import MockEvmConnector from '../../src/MockConnector';
+
+describe('Evm connect', () => {
```
ErnoW, line 5:
```suggestion
describe('transferErc20', () => {
```

ErnoW (changes requested):
Looks good to me.
One issue with the naming.

Also, I think we should account for error codes that we might have, (for example of possible errors see: https://www.blocknative.com/blog/ethereum-transaction-errors#ErrInsufficientFunds) But we probably should update the mockConnector to easily account for this.

We might need to do a bit of research to see what possible values we can get back as transaction/transactionReceipt. I think the errors are derived from that content. then we need a better way to set the TX/TxResponse in the mockConnector. 

Let me know what you think. We can do that outside the scope of this PR tough.

PR #510, opened 2022-07-18
feat: add Auth package and parse-server demo

ErnoW:
chore: remove logs

fix: update statement

docs: update changeset

chore: cleanup

---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

- Add Moralis.Auth package to handle authentication via the Moralis Authentication Api
- Add parse-server demo project


demos/demo-parse-server-auth/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+Now your server is running locally with the following endpoints:
+
+- Parse Server: `localhost:1337/server` (or any other port/endpoint you set in `.env`)
+- Parse Dashboard: `localhost:1337/dashboard` (or any other port you set in `.env`)
+- API: `localhost:1337/api` (or any other port you set in `.env`)
+
+### Run the client
+
+Go to the client directory: `cd client/public`
```
b4rtaz, line 21: This could be simplified by using `app.use(express.static('public'))` in the express server. Now we have two stage process to run the demo. It could be one stage.

demos/demo-parse-server-auth/README.md
```diff
@@ -0,0 +1,27 @@
+# Demo parse-server Auth
+
+## Run locally
+
+### Run the server
+
+Go to the server directory: `cd server`
+
+1. Run a mongo-db via `yarn dev:db-start` or run your a mongo-db in any other way
+2. Copy `.env.exampl` to `.env` and fill in the values
```
b4rtaz, line 10: Typo, should be: `.env.example`

packages/auth/src/methods/verify.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export interface SignMessageEvmOptions {
+  message: string;
+  signature: string;
+  network: 'evm';
+}
+
+export type SignMessageOptions = SignMessageEvmOptions;
+
+const makeEvmVerify = ({ network, ...options }: SignMessageEvmOptions) => {
```
sogunshola, line 12: `network` does not seem to be used here

packages/auth/src/AuthApi/AuthApi.ts
```diff
@@ -0,0 +1,7 @@
+import { completeChallenge } from './resolvers/completeChallenge';
+import { initializeChallenge } from './resolvers/initializeChallenge';
+
+export const AuthApi = {
+  requestMessage: initializeChallenge.fetch,
+  verify: completeChallenge.fetch,
+};
```
b4rtaz, line 7: These methods use a default instance of `MoralisCore` and there is no possibility to change it. This limits a user's use-cases. With that architecture I can't create a multi-tenancy architecture (one server/one instance of nodejs, multiple moralis API keys for different clients).

demos/demo-parse-server-auth/server/src/config.ts
```diff
@@ -0,0 +1,22 @@
+import * as dotenv from 'dotenv';
+
+dotenv.config();
+import { cleanEnv, num, str, bool } from 'envalid';
```
b4rtaz, line 4: Can this import be in imports section at the top?

packages/auth/src/AuthApi/resolvers/AuthApiResolver.ts
```diff
@@ -0,0 +1,172 @@
...
+  apiToResult: (result: ApiResult, params: Params) => AdaptedResult;
+  resultToJson: (result: AdaptedResult) => JSONResult;
+  parseParams: (params: Params) => ApiParams;
+  method?: Method;
+  bodyParams?: readonly (keyof Params)[];
+  bodyType?: BodyType;
+  name: string;
+}
+
+export class AuthApiResolver<ApiParams, Params, ApiResult, AdaptedResult, JSONResult> {
```
b4rtaz, line 28: A see here a lot `protected` fields. Why `protected` instead `private`? This class is not inherited anywhere.

sogunshola (commented):
Awesome work 🔥. few comments.

demos/demo-parse-server-auth/server/src/auth/MoralisAuthAdapter.ts
```diff
@@ -0,0 +1,43 @@
...
+    .catch(() => {
+      throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'Moralis auth failed, invalid data');
+    });
+}
+
+export function validateAppId(appIds, authData, options) {
+  return Promise.resolve();
+}
+
+export default {
```
b4rtaz, line 40: Why `export default` is here, if you have exported functions above?

demos/demo-parse-server-auth/server/src/auth/authController.ts
```diff
@@ -0,0 +1,38 @@
...
+      signature,
+    });
+
+    res.status(200).json({ user });
+  } catch (err) {
+    next(err);
+  }
+}
+
+export default {
```
b4rtaz, line 35: Why `export default` is here, if you have exported functions above?

demos/demo-parse-server-auth/server/src/auth/MoralisAuthAdapter.ts
```diff
@@ -0,0 +1,43 @@
+import Moralis from 'moralis';
+// Note: do not import Parse dependency. see https://github.com/parse-community/parse-server/issues/6467
+
+export function validateAuthData(authData, options) {
```
b4rtaz, line 4: This method is not used anywhere. So this demo doesn't test the verification part?

packages/core/src/Error/ErrorCode.ts
```diff
@@ -78,10 +78,19 @@ export enum UtilsErrorCode {
   NOT_IMPLEMENTED = 'U9000',
 }
 
+export enum AuthErrorCode {
```
b4rtaz, line 81: Here `Auth` domain logic leaks to `Core`. The same refers to `MoralisNetworkError`, `MoralisServerError` etc. Maybe not now, but that could be moved to appropriate packages.

b4rtaz (changes requested):
A few suggestions.

packages/auth/src/methods/verify.ts
ErnoW: Correct, we destructure it, so that we don't send it to requestMessage where the rest options are spread. I'll add eslint-disable-next-line @typescript-eslint/no-unused-vars to disable the lint warning

packages/auth/src/AuthApi/AuthApi.ts
ErnoW: Yes, I'm aware of that downside. Will revisit this when we have merged the refactoring regarding the config/circular-dependencies, and when we work on the SolApi.

We will need to revisit this anyway to reuse apiLogic between AuthApi, EvmApi and SolApi. For now it is a similar copy of EvmApi.

packages/auth/src/MoralisAuth.ts
```diff
@@ -0,0 +1,18 @@
+import core, { BaseModule } from '@moralisweb3/core/lib';
+import { makeRequestMessage, RequestMessageOptions } from './methods/requestMessage';
+import { makeVerify, SignMessageOptions } from './methods/verify';
+
+export class MoralisAuth extends BaseModule {
```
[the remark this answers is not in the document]
ErnoW: Yes, preferably we would have a package `ApiUtils` with generalised resolvers/adapters that we can re-use.

Lets do this before/when we start working on SolApi

packages/auth/src/AuthApi/resolvers/AuthApiResolver.ts
ErnoW: It is a copy of the Resolver of EvmApi, where it is extended into a PaginatedResolver.
I removed the PaginatedResolver in this module, as we don't need it.

But when we add a generalized Resolver that can be used in EvmApi, AuthApi and SolApi, then we will use the protected fields again, as it will need a PaginatedResolver.

demos/demo-parse-server-auth/server/src/auth/MoralisAuthAdapter.ts
ErnoW: Removed the individual exports

demos/demo-parse-server-auth/server/src/auth/authController.ts
ErnoW: Removed default export, only exporting named exports here now.

demos/demo-parse-server-auth/server/src/auth/MoralisAuthAdapter.ts
ErnoW: This is how custom authentication adapters work in Parse.
You need to provide a module that has the validateAuthData and validateAppId methods.

Then we provide this module when creating the parse server in `./src/parseServer`

Since we only care about validating authData, we only implement that method. But both are required.

For reference, this is how a firebase adapter looks like: https://github.com/parse-server-modules/parse-server-firebase-auth-adapter

packages/core/src/Error/ErrorCode.ts
ErnoW: Yes I agree.  A nice improvement would be to move the ErrorCodes to the corresponding packages.

I initially had them in `core` as it is a nice reference file where you can see all errors. But ideally we would move them to the correct packages and generate a docs page that combines all ErrorCodes from the packages.

PR #517, opened 2022-07-19
feat(sol-api): solana API, early stage.

b4rtaz:
---
name: Solana API
---

### Checklist


packages/solApi/src/resolvers/account/getNFTs.ts
```diff
@@ -0,0 +1,41 @@
...
+  resultToJson: (data) => {
+    return data.map((nft) => {
+      return {
+        associatedTokenAddress: nft.associatedTokenAddress.toJSON(),
+        mint: nft.mint.toJSON(),
+      };
+    });
+  },
+  parseParams: (params: Params): ApiParams => ({
+    network: SolNetwork.create(params.network).network,
```
sogunshola, line 38: I think it would be nice if we can make the network optional and default to `mainnet` if no network param is passed

sogunshola (commented):
Left one question otherwise, everything looks good 👍🏾

packages/solApi/src/resolvers/account/getNFTs.ts
b4rtaz: Good idea!

packages/solApi/src/resolvers/SolNetworkResolver.ts
```diff
@@ -0,0 +1,12 @@
+import { MoralisCoreProvider, SolNetwork, SolNetworkish, SolNetworkName } from '@moralisweb3/core';
+import { SolApiConfig } from '../config/SolApiConfig';
+
+export class SolNetworkResolver {
```
sogunshola, line 4: in the API package, there is a file `packages/api/src/utils/resolveParams`. This has a similar function but for resolving default chain.

Which do you think makes more sense.

- Move `SolNetworkResolver` to that file in the `api` package or
- Move `resolveDefaultChain` to the evmApi package

I think doing one of these 2 will give a better structure.

b4rtaz: I'll move `resolveDefaultChain` to the evmApi package. Good point!

.github/workflows/main.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         id: cache
         with:
           path: '**/node_modules'
-          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}_C1
```
ErnoW, line 33: What is the reason for this? Is it to invalidate the current cache?

demos/test-node/src/EvmApi.ts
```diff
@@ -0,0 +1,41 @@
+import Moralis from 'moralis';
+import { smokeTest } from './Tester';
```
ErnoW, line 2: Out of scope for this PR but I think most of the functionalities should be tested as:
- unit test in the specified package
- integration tests in the integrations package

This app should more be a test app for nodeJs, where we can test functionalities easily. We probably could change this to a proper demo app.

Or do you see any additional benefits of running tests here as well?

packages/core/src/dataTypes/SolNative.ts
```diff
@@ -0,0 +1,71 @@
...
+      throw new MoralisCoreError({
+        code: CoreErrorCode.INVALID_ARGUMENT,
+        message: `Not supported Solana unit: ${unit}`,
+      });
+    }
+    return parseFixed(value.toString(), decimal);
+  }
+
+  // TODO: we cannot share an internal instance of BigNumber, because BigNumber is located in @ethersproject.
+  // We should use some class from Solana JS stack. For example @solana/web3.js uses BN.js.
```
ErnoW, line 36: I agree that we should not depend on EthersJs usage of BigNumber.

Ideally we should use the BigInt type. I know that EthersJs is also shifting to that and deprecate the use of BigNumber.

Will that work with Solana, do you think?

Otherwise we can use BN.js for Solana, as @solana/web3.js is the main SDK for solana integrations

ErnoW: This should also remove the dependency of @ethersproject in the Sol datatypes which is good.

We will split up the core module in:
- Core
- EvmUtils / EvmData
- SolUtils / SolData

So it would be perfect if the solana packages have no dependencies on ethersJs

packages/core/src/dataTypes/SolNetwork.ts
```diff
@@ -0,0 +1,36 @@
+import { MoralisData, MoralisDataFormatted } from './abstract';
+
+export type SolNetworkName = 'mainnet' | 'devnet';
```
ErnoW, line 3: I think we should have a consistent naming convention. Currently we have
Network = evm/solana/elrond etc.
Chain for Evm is specified by the chainId

Now Solana has the concept of clusters https://docs.solana.com/clusters which they describe the difference between mainnet and devnet

This is something that affects Moralis in broader terms, and will leave it out of scope of this PR. And I think Network is currently correct as it is implemented that way in the Api as well.

demos/test-node/src/EvmApi.ts
b4rtaz: I agree. But here we have smoke tests. I would say, that is something like a playground. I needed some place to run an implementation with the live data.

packages/core/src/dataTypes/SolNetwork.ts
b4rtaz: 👍

packages/core/src/dataTypes/SolAddress.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  private static parse(address: string): string {
+    // TODO: add address validation
+    return address;
+  }
+
+  public get lowercase() {
+    return this.address.toLowerCase();
+  }
```
ErnoW, line 17: We should remove this. 
Solana addresses are different than EVM in that they are case-sensitive

b4rtaz: Removed.

ErnoW (approved):

PR #518, opened 2022-07-20
test(api): api package test + few adjustments

sogunshola:
---
name: API package test
---

### Checklist

### Issue Description

### DIFF
--- a/packages/api/src/test/setup.ts
+++ b/packages/api/src/test/setup.ts
@@ -1,0 +1,122 @@
+import { MoralisApi } from './../MoralisApi';
+import { ApiModule, EvmAddress, EvmAddressish, EvmChainish, MoralisCore, MoralisCoreProvider } from '@moralisweb3/core';
+import { setupServer } from 'msw/node';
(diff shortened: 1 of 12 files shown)


packages/api/src/test/setup.ts
```diff
@@ -0,0 +1,122 @@
...
+  public setup() {
+    //
+  }
+
+  public start(): void | Promise<void> {
+    // Nothing...
+  }
+}
+
+const mockServer = setupServer(mockEndpointWeights, mockGetContractEvents);
```
b4rtaz, line 104: I think this is not the best way how it could be tested. We already have integration tests, so I think we don't need a next integration test package. You can here mock responses by `jest.spyOn(axios, 'request')` instead using the `msw` server.

b4rtaz (changes requested):

packages/api/src/test/setup.ts
sogunshola: good suggestion. on it

b4rtaz (changes requested):

PR #521, opened 2022-07-21
Move evm logic to evm utils

ErnoW:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

- Moved all Evm dataTypes to `evm-utils`
- Moved register of config of `formatEvmAddress` and `formatEvmChainId` to `evm-utils`
- Moved register of config of `defaultEvmApiChain` to `evm-api`

Other fixes/improvements
- Naming of `setupEvmApi`
- Config of eslint was broken (syntax error)
- Fixed versioning of newly created packages
- cache format function in nx
- Fix prettier versioning as devDependency
- Fix building of test-node demo app by using symLinked packages when building


packages/core/package.json
```diff
@@ -29,10 +29,8 @@
     "typescript": "^4.5.5"
   },
   "dependencies": {
-    "@ethersproject/address": "^5.6.0",
     "@ethersproject/bignumber": "^5.6.0",
     "@ethersproject/units": "^5.6.0",
```
b4rtaz, line 33: Do we need these dependencies?

packages/evmUtils/src/config/EvmUtilsConfigSetup.ts
```diff
@@ -0,0 +1,14 @@
...
+import { EvmUtilsConfig } from './EvmUtilsConfig';
+
+export class EvmUtilsConfigSetup {
+  public static register(config: Config) {
+    if (!config.hasKey(EvmUtilsConfig.formatEvmAddress)) {
+      config.registerKey(EvmUtilsConfig.formatEvmAddress);
+    }
+
+    if (!config.hasKey(EvmUtilsConfig.formatEvmChainId)) {
+      config.registerKey(EvmUtilsConfig.formatEvmChainId);
```
b4rtaz, line 11: I think this approach is wrong. This approach hides modules usage errors. You should get an error, if you use the module system in wrong way.

packages/evmUtils/src/test/setup.ts
```diff
@@ -0,0 +1,11 @@
+import { MoralisCoreProvider } from '@moralisweb3/core';
+import { MoralisEvmUtils } from '../MoralisEvmUtils';
+
+export const setupEvm = () => {
```
b4rtaz, line 4: Should be `setupEvmUtils()`.

b4rtaz (changes requested):
Just 3 remarks. Good job.

packages/core/package.json
ErnoW: Only as long as it is being used by the Sol datatypes in core.

b4rtaz (approved):

PR #550, opened 2022-08-02
feat: add tracking for api calls

ErnoW:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description


packages/core/src/MoralisCore.ts
```diff
@@ -27,6 +28,8 @@ export class MoralisCore {
 
   public readonly name = MoralisCore.moduleName;
 
+  public readonly libVersion = LIB_VERSION;
```
b4rtaz, line 31: This could be `static`.

packages/apiUtils/src/resolvers/getCommonHeaders.ts
```diff
@@ -0,0 +1,7 @@
+export type Headers = { [key: string]: string };
+
+export const getCommonHeaders = (libVersion: string): Headers => ({
+  'x-moralis-platform': 'JS SDK',
+  'x-moralis-platform-version': libVersion,
```
b4rtaz, line 5: If `Core.libVersion` would be `static`, here you don't need to pass `libVersion`.

b4rtaz (changes requested):

packages/apiUtils/src/resolvers/EndpointResolver.test.ts
```diff
@@ -83,4 +84,32 @@ describe('ApiResolver', () => {
...
+        throw new Error('x-moralis-platform-version is not set to the correct');
+      }
+
+      if (!options.headers['x-moralis-build-target']) {
+        throw new Error('x-moralis-build-target is not set');
+      }
+
+      if (options.headers['x-api-key'] !== MOCK_API_KEY) {
+        throw new Error('x-api-key is not set to the correct');
+      }
```
b4rtaz, line 106: It's better to use `expect()` to assert in tests. It doesn't matter that, that expectation is inside in a mock implementation.

```
expect(options.headers['x-moralis-platform-version']).toBe(MoralisCore.libVersion);
...
```

b4rtaz (changes requested):

packages/apiUtils/src/resolvers/EndpointResolver.test.ts
```diff
@@ -83,4 +84,32 @@ describe('ApiResolver', () => {
...
+
+      return {};
+    });
+
+    const response = await resolver.fetch({});
+
+    expect(response).toBeDefined();
```
b4rtaz, line 113: Here you need `expect(mockRequest).toHaveBeenCalledTimes(1);`. Because you want to prove that, asserts in the mock function was called. Currently this test may pass, without asserts in the mock function.

b4rtaz (changes requested):

packages/apiUtils/src/resolvers/EndpointResolver.test.ts
ErnoW: It might be more difficult to read, as I prefer to use the arrange-act-assert method.

I'll try a different solution

PR #566, opened 2022-08-09
feat: firebase proxy demo + firebase auth demo.

b4rtaz:
---
name: Firebase Proxy Demo
---

### Checklist


sogunshola (commented):
This looks good to me. Just one question.

can the functions be auto generated for all the api endpoints?

b4rtaz:
> can the functions be auto generated for all the api endpoints?

I've just checked it and it's possible.

```ts
export const api: { [functionName: string]: functions.HttpsFunction } = {};

for (let i = 0; i < 10; i++) {
  api[`func_${i}`] = functions.https.onCall(
    limiter(async (req, res) => {
      return 'test';
    }));
}
```

The above code creates endpoints like:

```
✔  functions[us-central1-api-func_0]: http function initialized (http://localhost:5001/moralis-x3/us-central1/api-func_0).
✔  functions[us-central1-api-func_1]: http function initialized (http://localhost:5001/moralis-x3/us-central1/api-func_1).
✔  functions[us-central1-api-func_2]: http function initialized (http://localhost:5001/moralis-x3/us-central1/api-func_2).
✔  functions[us-central1-api-func_3]: http function initialized (http://localhost:5001/moralis-x3/us-central1/api-func_3).
...
```

So we could create endpoints dynamical. But I think we shouldn't create a proxy for all API endpoints, because that would create a disorder in the Firebase admin. If we add some way to filter methods, that would solve this problem.

I will try to figure out this.

Good point @sogunshola!

ErnoW (changes requested):
As discussed, lets move the proxy utilities for now to the demo (instead of a separate package). And we can generalise a proxy generator more generically. Other than that it looks good to me.

We will need to do some polishing to make all the demos a bit more uniform, but I'll create a separate issue for this (mainly regarding eslinting, prettier/formatting rules etc.). It should be a uniform experience.

ErnoW (approved):

PR #567, opened 2022-08-09
Feat/express proxy

sogunshola:
---
name: Express app api proxy using axios
---

### Checklist

### Issue Description


demos/parse-server/src/api/evmProxy.ts
```diff
@@ -0,0 +1,43 @@
...
+import express from 'express';
+import axios from 'axios';
+import config from '../config';
+
+const evmProxyRouter = express.Router();
+
+const descriptors = Moralis.EvmApi.endpoints.getDescriptors();
+
+for (const descriptor of descriptors) {
+  evmProxyRouter.route(`/${descriptor.name}`).post(async (req, res) => {
```
b4rtaz, line 11: I think here you could put `.urlPattern` as an argument of the `.route(...)` method. **In this way, the proxy would have the same URL patterns as Moralis API has.**

```js
for (const descriptor of descriptors) {

  async function requestHandler(req, res) {
    // ...
  }
  const urlPattern = descriptor.urlPattern.replace(/\{/g, ':').replace(/\}/g, ''); // This converts an "universal" route format to express.js format.
  const routeBuilder = router.route('proxy' + urlPattern);
  switch (descriptor.method) {
    case 'get':
      routeBuilder.get(requestHandler);
      break;
     case 'post':
      routeBuilder.post(requestHandler);
      break;
}
```

<pre>
https://x.com/proxy/<strong>account/{network}/{address}/balance</strong>
https://x.com/proxy/<strong>account/{network}/{address}/nft</strong>
...
</pre>

We should keep the same type of methods here (GET, POST).

b4rtaz (changes requested):

demos/parse-server/public/script.js
```diff
@@ -28,6 +43,18 @@ const verifyMessage = (message, signature) =>
     network: 'evm',
   });
 
+// proxy calls
+const web3apiVersion = () => {
+  handleEvmProxyCall('web3ApiVersion');
```
b4rtaz, line 48: After [this](https://github.com/MoralisWeb3/Moralis-JS-SDK/pull/567/files#r941763742) remarks, here should be `handleEvmProxyCall('GET', 'web3/version');`.

demos/parse-server/public/script.js
```diff
@@ -28,6 +43,18 @@ const verifyMessage = (message, signature) =>
...
+};
+
+const getNativeBalance = () => {
+  handleEvmProxyCall('getNativeBalance', {
```
b4rtaz, line 52: After [this](https://github.com/MoralisWeb3/Moralis-JS-SDK/pull/567/files#r941763742) remarks, here should be `handleEvmProxyCall('GET', 'info/endpointWeights');`.

b4rtaz (changes requested):

demos/parse-server/src/api/evmProxy.ts
```diff
@@ -0,0 +1,43 @@
...
+        },
+      })
+      .then((response) => res.send(response.data))
+      .catch((error) => {
+        res.status(500).send(error);
+      });
+  });
+}
+
+export default evmProxyRouter;
```
b4rtaz, line 43: Maybe this router should be in the own package (`@moralisweb3/expressjs`)? Because this logic would be exactly the same for the parse-server.

b4rtaz (changes requested):

demos/parse-server/src/apiRouter.ts
```diff
@@ -1,6 +1,8 @@
 import express from 'express';
+import evmProxyRouter from './api/evmProxy';
 import { authRouter } from './auth/authRouter';
 
 export const apiRouter = express.Router();
 
 apiRouter.use('/auth', authRouter);
+apiRouter.use('/evm-proxy', evmProxyRouter);
```
b4rtaz, line 8: `/evm-proxy` -> `/evm-api-proxy`

b4rtaz (changes requested):

b4rtaz:
Sorry for multiple requests. I've already finished. 😄

b4rtaz:
I think the below files:

```
demos/express-proxy/src/api/*
demos/express-proxy/src/apiRouter.ts
demos/express-proxy/src/middlewares/*
```

You can extract to own package: `@moralisweb3/express`. The demo app will use this package.

By this, you can easily attach proxy to `demos/express-proxy` and `demos/parse-server` and keep the code in single place.

b4rtaz (changes requested):

demos/express-proxy/src/api/solanaProxy.ts
```diff
@@ -0,0 +1,51 @@
...
+        },
+      })
+      .then((response) => res.send(response.data))
+      .catch((error) => {
+        errorHandler(error, req, res, next);
+      });
+  });
+}
+
+export default solanaProxyRouter;
```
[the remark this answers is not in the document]
sogunshola: Nice idea, will do that

demos/express/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@moralisweb3/express",
```
ErnoW, line 2: Similar to what we did with the firebase demo: lets not create a separate package to create proxies. We need to have a generic package that can be used for any backend that creates procies. As this approach will result in maintaining a large amount of integrations that are very opinionated (devs might want to implement their own systems, so we need to be flexible).

demos/express/src/middlewares/rateLimiter.ts
```diff
@@ -0,0 +1,51 @@
+import { createClient } from 'redis';
```
ErnoW, line 1: For a simple express demo I would say that rate-limiting is a bit much, and we could keep it as simple as possible.

For the parse-demo we should include it for sure I would say. As we can use that demo to migrate people over from v1 to a self hosted parse-server in v2

ErnoW (changes requested):

demos/express/package.json
sogunshola: 👍🏾 Included the proxy generator in proxy package now

demos/express/src/middlewares/rateLimiter.ts
b4rtaz: I found the [express-rate-limit](https://github.com/nfriedly/express-rate-limit) package. It looks like easy to use and it doesn't require redis. By default it uses the memory store.

ErnoW:
Currently I see references to both `redis` and `express-rate-limit`. 
Can we only use express-rate-limit in this demo?

ErnoW:
Other than that, this demo looks good to me

b4rtaz (approved):

b4rtaz (changes requested):
We need the `express-rate-limit` package instead pure redis.

demos/parse-server/.env.example
```diff
@@ -13,4 +13,5 @@ MASTER_KEY = 'MASTER_KEY'
 APPLICATION_ID = 'APPLICATION_ID'
 SERVER_URL = 'http://localhost:1337/server'
 
-ALLOW_INSECURE_HTTP = 'true'
\ No newline at end of file
+ALLOW_INSECURE_HTTP = 'true'
+REDIS_URL='redis://localhost:6379'
```
b4rtaz, line 17: Is this still needed (`REDIS_URL`)?

sogunshola: No. I just removed it

b4rtaz (approved):

ErnoW (approved):

PR #591, opened 2022-08-18
Prettier fixes

ErnoW:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description
- Nested prettierignore files do not work properly with Vscode and thus was conflicting (local formatting vs CI check)
- Inconsistencies in some demo apps regarding formatting have been removed (now we only use the prettier root)
- Removed some eslint rules for the firebase app as it was conflicting with formatting rules. We will create a eslint rule-set that we will use in all the demo apps as well


demos/firebase-auth/functions/package.json
```diff
@@ -23,9 +23,6 @@
   "devDependencies": {
     "@typescript-eslint/eslint-plugin": "^5.12.0",
     "@typescript-eslint/parser": "^5.12.0",
-    "eslint": "^8.9.0",
-    "eslint-config-google": "^0.14.0",
-    "eslint-plugin-import": "^2.25.4",
```
b4rtaz, line 28: I think this is not the best approach. **By this we give users a demo with no eslint**. After I copy the demo folder to my local machine, I have a nice demo but eslint doesn't work. So we don't provide the best programming practices. I believe that, some users will build an own product on this demo.

```
b4rtaz@macbook functions % npm run lint

> firebase-auth-functions@1.0.0 lint
> eslint --ext .js,.ts .


Oops! Something went wrong! :(

ESLint: 8.22.0

ESLint couldn't find a configuration file. To set up a configuration file for this project, please run:

    npm init @eslint/config
```

b4rtaz (changes requested):

demos/firebase-auth/functions/package.json
b4rtaz: Basically the `<root_of_repo>/.eslintrc.js` file has only 1 393 bytes, we could duplicate it be each demo. So that means, we should have one `.eslintrc.js` for SDK packages, and many `.eslintrc.js` for each demo.

ErnoW: What I want to do is create a new package that contains the eslint rules, that way we can have an eslintrc in every project as well, that is conform with our own lint styling. Users can then easily opt-out or change the rules.

I removed it for now, as it was conflicting with prettier. But I do agree that we need a simple eslint in every project

ErnoW: This is now fixed with the introduction of the eslintConfig package

packages/core/src/controllers/AxiosRetry.ts
```diff
@@ -16,6 +16,8 @@ export class AxiosRetry {
   ): Promise<AxiosResponse<Response>> {
     for (let attempt = 1; ; attempt++) {
       try {
+        // TODO: refactor to fix this error
```
b4rtaz, line 19: I think here refactor is NOT necessary. This rule refers to cases where you can put all promises into `Promise.all()`.

> When Not To Use It
> 
> In many cases the iterations of a loop are not actually independent of each-other. For example, the output of one iteration might be used as the input to another.

Source: https://eslint.org/docs/latest/rules/no-await-in-loop#when-not-to-use-it

b4rtaz (changes requested):

packages/core/src/controllers/AxiosRetry.ts
ErnoW: Ah yes you're right, it does not apply in this situation

PR #605, opened 2022-08-23
docs: create examples overview

ErnoW:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

Generates an example overview page with all the demo projects


scripts/buildExamples.js
```diff
@@ -0,0 +1,264 @@
...
+// Compression level vor arciving the code to .zip
+const COMPRESSION_LEVEL = 9;
+
+// References to our folder names
+const DEMO_FOLDER_DIR = 'demos';
+const ASSETS_FOLDER_DIR = 'assets';
+const OUT_FOLDER_DIR = 'pages';
+const DOCS_FOLDER_DIR = 'docs';
+// Reference to the output example folder inside the docs folder
+const OUT_EXAMPLE_FOLDER_DIR = 'examples';
```
b4rtaz, line 26: We should avoid giving a few names for the same thing. We have `demos`, here I see `examples`. So we have two names describe the same. I think here should be `demos-release`, `demo-archives` or similar.

https://enterprisecraftsmanship.com/posts/ubiquitous-language-naming/

b4rtaz (changes requested):

scripts/buildExamples.js
ErnoW: Ah yeah, the names make no sense at the moment after all the refactors.
I think it is better now.

b4rtaz (approved):

PR #616, opened 2022-08-25
docs: EvmChain

sogunshola:
add technical docs for EvmChain

---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description


packages/evmUtils/src/dataTypes/EvmChain/EvmChain.ts
```diff
@@ -12,14 +12,26 @@ import {
...
- * hex-string ("0x1"), or any suppored ChainName
+ * @category DataType
  */
 export class EvmChain implements MoralisData, EvmChainable {
+  /**
+   * Create a new instance of EvmChain from any valid address input.
+   *
+   * @example
+   * ```ts
+   * const chain = EvmChain.create(1)
```
ErnoW, line 32: Lets add also an example with hex decimal string as input here

packages/evmUtils/src/dataTypes/EvmChain/EvmChain.ts
```diff
@@ -12,14 +12,26 @@ import {
...
- * A new instance can be created via `EvmChain.create(chain)`, where the provided chain can be a valid chain number (1),
- * hex-string ("0x1"), or any suppored ChainName
+ * @category DataType
  */
 export class EvmChain implements MoralisData, EvmChainable {
+  /**
+   * Create a new instance of EvmChain from any valid address input.
+   *
+   * @example
+   * ```ts
```
ErnoW, line 31: Does it work better when we specify `ts` here? 🤔

packages/evmUtils/src/dataTypes/EvmChain/EvmChain.ts
```diff
@@ -80,9 +49,7 @@ export class EvmChain implements MoralisData, EvmChainable {
     this._chainlistData = chainList.find((chainData) => chainData.chainId === this.decimal) ?? null;
   }
 
-  /**
-   * Getter to return _chainlistData and throws an error if it is not defined
-   */
+  // Getter to return _chainlistData and throws an error if it is not defined
```
ErnoW, line 52: Why remove the code-block here?

packages/evmUtils/src/dataTypes/EvmChain/EvmChain.ts
```diff
@@ -91,22 +58,116 @@ export class EvmChain implements MoralisData, EvmChainable {
...
+  /**
+   * Compares if 2 chains are equal, based on the chainId
+   *
+   * @param chainA - The first chain to compare
+   * @param chainB - The second chain to compare
+   *
+   * @returns true if the chains are equal, false otherwise
+   * @example
+   * ```ts
+   * EvmChain.equals("0xfB6916095ca1df60bB79Ce92cE3Ea74c37c5d359", "0xfb6916095ca1df60bb79ce92ce3ea74c37c5d359")
```
ErnoW, line 70: These are not chains

packages/evmUtils/src/dataTypes/EvmChain/EvmChain.ts
```diff
@@ -91,22 +58,116 @@ export class EvmChain implements MoralisData, EvmChainable {
...
+      return this.hex;
+    }
+
+    return assertUnreachable(formatStyle);
+  }
+
+  /**
+   * Displays the chain name or hex-string representation of the chain
+   *
+   * @example chain.display() // "Ethereum"
```
ErnoW, line 117: Don't think this is true, as it returns the hex as well (and name can be null)

packages/evmUtils/src/dataTypes/EvmChain/EvmChain.ts
```diff
@@ -91,22 +58,116 @@ export class EvmChain implements MoralisData, EvmChainable {
...
+   */
+  display() {
+    return this.name ? `${this.name} (${this.hex})` : this.hex;
+  }
+
+  getExplorerPath(value: { block: string } | { transaction: string } | { account: string } | { erc20: string }) {
```
ErnoW, line 123: Add description for this function as well

packages/evmUtils/src/dataTypes/EvmChain/EvmChain.ts
```diff
@@ -127,88 +188,169 @@ export class EvmChain implements MoralisData, EvmChainable {
...
-    return EvmChain.create(chainA)._value === EvmChain.create(chainB)._value;
+  get name() {
+    return this._getChainlistData()?.name;
   }
 
   /**
-   * Compares if the current chain is equal to the provided chain, based on the chainId
+   * Returns the currency of the chain
+   *
+   * @example chain.currency // "ETH"
```
ErnoW, line 201: Please check the typescript definition, the return type is different (and can  be undefined as well)

packages/evmUtils/src/dataTypes/EvmChain/EvmChain.ts
```diff
@@ -127,88 +188,169 @@ export class EvmChain implements MoralisData, EvmChainable {
...
-  display() {
-    return this.name ? `${this.name} (${this.hex})` : this.hex;
+  /**
+   * Returns ETHEREUM chain
+   *
+   * @example EvmChain.ETHEREUM // EvmChain
+   */
+  public static get ETHEREUM() {
+    return EvmChain.create(1);
   }
```
ErnoW, line 238: All static getters / enums, should be at the top. Like we have in EvmAddress.
I just updated this in our internal docs, as I forgot to mention this.

ErnoW (changes requested):

packages/evmUtils/src/dataTypes/EvmChain/EvmChain.ts
sogunshola: I think so. because it a `.md` file now, when I preview it without ```ts I get a black and white code block

packages/evmUtils/src/dataTypes/EvmChain/EvmChain.ts
sogunshola: It's a private function and I am not sure if we want to document private functions. Do we? 🤔

b4rtaz: I think no.

ErnoW: It should not display in the docs as we have this setup. Only public methods should be documented:
```
  "visibilityFilters": {
    "protected": false,
    "private": false,
    "inherited": false,
    "external": false,
    "@alpha": false,
    "@beta": false
  },
```

But if we add a comment to a function, class etc, then lets keep it in a uniform style with code-blocks. As this makes it easier to refactor to public (if needed later on), or other integrations as the vscode autocomplete.

packages/evmUtils/src/dataTypes/EvmChain/EvmChain.ts
```diff
@@ -12,65 +12,161 @@ import {
 } from '@moralisweb3/core';
 import { EvmUtilsConfig } from '../../config/EvmUtilsConfig';
 
+/**
+ * This can be any valid EVM chain in decimal or hex.
+ * @example "3"
```
ErnoW, line 17:
```suggestion
 * @example 3
```
Should not be a string

packages/evmUtils/src/dataTypes/EvmChain/EvmChain.ts
```diff
@@ -158,32 +238,25 @@ export class EvmChain implements MoralisData, EvmChainable {
     return assertUnreachable(formatStyle);
   }
 
+  /**
+   * Displays the chain hex-string representation of the chain and also the chain name if not null
+   *
+   * @example chain.display() // "Ethereum 0x1" | "0x1"
```
ErnoW, line 244:
```suggestion
   * @example chain.display() // "Ethereum (0x1)" | "0x1"
```

ErnoW (changes requested):

PR #625, opened 2022-08-29
docs: Erc20Value

sogunshola:
technical docs for Erc20Value

---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

### DIFF
--- a/packages/evmUtils/src/dataTypes/Erc20Value/Erc20Value.ts
+++ b/packages/evmUtils/src/dataTypes/Erc20Value/Erc20Value.ts
@@ -3,15 +3,33 @@ import { Erc20Token, Erc20Tokenish } from '../Erc20/Erc20';
 
 const EVM_ERC20_DEFAULT_DECIMALS = 18;
 
+/**
+ * {@link BigNumberish} type for the amount of tokens
+ */
 export type Erc20ValueInputAmount = BigNumberish;
+
+/**
+ * This is a number or a string that represents the decimals of tokens
+ */
 export type Erc20ValueInputDecimals = number | string;
+
+/**
+ * Valid input for a new Erc20Value instance.
+ * This can be an existing {@link Erc20Value} or a valid {@link Erc20ValueInputAmount} object
+ */
 export type Erc20Valueish = Erc20ValueInputAmount | Erc20Value;
 
+/**
+ * This is the return type of the processed Erc20Value
+ */
 export type Erc20ValueData = {
   amount: BigNumber;
   decimals: number;
 };
 
+/**
+ * The options for the Erc20Value class
+ */
 export type Erc20Options = {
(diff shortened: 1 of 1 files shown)


packages/evmUtils/src/dataTypes/Erc20Value/Erc20Value.ts
```diff
@@ -110,14 +149,56 @@ export class Erc20Value implements MoralisData {
...
   toJSON() {
     if (this.token) {
       return { value: this.value, token: this.token.toJSON() };
     }
     return { value: this.value };
   }
+
+  /**
+   * @returns the token decimals
+   * @example value.decimals;
```
ErnoW, line 175: Same as previous PR, please add an example return value, otherwise it has little added value

packages/evmUtils/src/dataTypes/Erc20Value/Erc20Value.ts
```diff
@@ -59,49 +94,53 @@ export class Erc20Value implements MoralisData {
...
-
-    return new Erc20Value(value, options);
-  }
-
+  /**
+   * Compares two Erc20Valueish instances.
+   * @param valueA - The first value to compare
+   * @param valueB - The second value to compare
+   * @returns True if the values are equal
+   * @example Erc20Value.equals(valueA, valueB);
```
ErnoW, line 102: Please provide a better example with real values. For example, it's important for users to know that 
{value: 10000, decimals:4} equals {value 1000, decimals:3}

ErnoW (changes requested):

packages/evmUtils/src/dataTypes/Erc20Value/Erc20Value.ts
```diff
@@ -59,49 +95,58 @@ export class Erc20Value implements MoralisData {
...
+  /**
+   * Compares two Erc20Valueish instances.
+   * @param valueA - The first value to compare
+   * @param valueB - The second value to compare
+   * @returns True if the values are equal
+   * @example
+   * ```ts
+   * const valueA = Erc20Value.create(1000, { decimals: 3 });
+   * const valueB = Erc20Value.create(10000, { decimals: 4 });
+   * Erc20Value.equals(valueA, valueB); // false
```
ErnoW, line 107: This should return `true`, did it return `false` for you? That should not be the case and it is covered by a unit-test. Because 1.000 equals 1.0000.

Please check this.

ErnoW (changes requested):

packages/evmUtils/src/dataTypes/Erc20Value/Erc20Value.ts
sogunshola: Ah you are right. just tested now

PR #626, opened 2022-08-29
docs: BigNumber

sogunshola:
technical docs for BigNumber datatype

---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description


packages/core/src/dataTypes/BigNumber/BigNumber.ts
```diff
@@ -7,56 +7,126 @@ import { BigNumberParser, BigNumberPrimitive } from './BigNumberParser';
...
+   */
   public toBigInt(): bigint {
     return this.value;
   }
 
+  /**
+   * Adds a BigNumber to current BigNumber instance.
+   * @param value - the BigNumberish to add
+   * @returns the result of the addition
+   * @example BigNumber.create(12).add(BigNumber.create(7));
```
ErnoW, line 55: Can we add multiple examples to these methods? As it is important for the user to know that they can use .add(1) as well, or .add("1000000000000000000")

ErnoW (changes requested):

packages/core/src/dataTypes/BigNumber/BigNumber.ts
ErnoW: Does it not work with this:
```
BigNumber.create(12).add(7)
BigNumber.create(12).add("1000000000000000000")
```
?
I think that is a better and easier example to give.

ErnoW (changes requested):

ErnoW (approved):

PR #636, opened 2022-08-30
Fix dependencies and eslint

ErnoW:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description
Fix a lot of warnings when running yarn install:
- Add missing peer dependencies
- Add correct eslint dependencies
- Add Eslint to every demo (and fix warnings)


ErnoW:
Note: this also removes the yarn.lock from the demo projects, as these are currently not updated due to the fact that it is managed by the monorepo (global yarn.lock). I created a separate issue to find a way to include lockfiles for each demo

b4rtaz (dismissed):

PR #641, opened 2022-08-31
docs: EvmNftTrade

sogunshola:
technical docs for nft trade datatype

---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

### DIFF
--- a/packages/evmUtils/src/dataTypes/EvmNftTrade/EvmNftTrade.ts
+++ b/packages/evmUtils/src/dataTypes/EvmNftTrade/EvmNftTrade.ts
@@ -1,12 +1,39 @@
-import { MoralisDataObject, maybe, BigNumber } from '@moralisweb3/core';
+import { MoralisDataObject, maybe, BigNumber, dateInputToDate } from '@moralisweb3/core';
 import { EvmAddress } from '../EvmAddress';
 import { EvmChain } from '../EvmChain';
 import { EvmNative } from '../EvmNative';
 import { EvmNftTradeInput, EvmNftTradeData } from './types';
 
+/**
+ * Valid input for a new EvmNftTrade instance.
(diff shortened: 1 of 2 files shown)


packages/evmUtils/src/dataTypes/EvmNftTrade/EvmNftTrade.ts
```diff
@@ -24,16 +51,19 @@ export class EvmNftTrade implements MoralisDataObject {
     blockNumber: BigNumber.create(data.blockNumber),
     price: EvmNative.create(data.price),
     transactionIndex: +data.transactionIndex,
+    blockTimestamp: dateInputToDate(data.blockTimestamp),
```
ErnoW, line 54: Please add a changeset (patch), as this is a feature change and should be documented for the user

ErnoW (changes requested):

packages/evmUtils/src/dataTypes/EvmNftTrade/types.ts
```diff
@@ -3,6 +3,26 @@ import { EvmAddressish, EvmAddress } from '../EvmAddress';
 import { EvmChain, EvmChainish } from '../EvmChain';
 import { EvmNativeish, EvmNative } from '../EvmNative';
 
+/**
+ * This can be any object with valid trade data.
+ * @example
+ * ```
```
ErnoW, line 9: I think this should be wrapped in an object, right?

ErnoW: Ah nevermind, it is, it looked weird at first glance. Because of the spacing. Let's see if doc-generation handles the spacing automatically. Otherwise we need to format this better

ErnoW (approved):

PR #657, opened 2022-09-07
Migration demo parse server

ErnoW:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

Create demo for migration of parse-server from hosted by moralis to self-hosted server, with client-side example (using v1 of the sdk)


demos/cli/tsconfig.json
```diff
@@ -10,7 +10,8 @@
     "esModuleInterop": true,
     "moduleResolution": "node",
     "lib": ["es2017"],
-    "composite": false
+    "composite": false,
+    "skipLibCheck": true
```
ErnoW, line 14: Added this to all packages as I was getting errors when building the app regarding types of react and react-dom. This also speeds up build time.

packages/solApi/src/MoralisSolApi.ts
```diff
@@ -30,11 +30,15 @@ export class MoralisSolApi extends ApiModule {
...
+  private readonly getBalance = this.endpoints.createFetcher(getBalance);
+
   public readonly account = {
-    getBalance: this.endpoints.createFetcher(getBalance),
+    getBalance: this.getBalance,
     getNFTs: this.endpoints.createFetcher(getNFTs),
     getPortfolio: this.endpoints.createFetcher(getPortfolio),
     getSPL: this.endpoints.createFetcher(getSPL),
+    // Support for old naming
+    balance: this.getBalance,
```
ErnoW, line 41: Since we build cloud functions based on the swagger.json for the parse-server demo, we need the reference to the `balance` method (that we named getBalance internally).
This won't be required anymore once we move to a generic way to generate the endpoints

demos/parse-server-migration/src/cloud/generated/evmApi.ts
```diff
@@ -0,0 +1,439 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
```
ErnoW, line 1: These are now generated via `yarn gen:cloud`. We can move to a more dynamic setup once we implement the generic proxy generatr. For now I think this is the best approach as it clearly shows the users what endpoints are available.

b4rtaz (commented):
My remarks have disappeared.

packages/evmApi/src/EvmApi.ts
```diff
@@ -357,12 +376,12 @@ export class MoralisEvmApi extends ApiModule {
     /**
      * @deprecated Replaced by `utils.endpointWeights()`.
      */
-    endpointWeights: () =>
+    endpointWeights: (_params?: unknown) =>
```
b4rtaz, line 379: Why we need this?

demos/parse-server-migration/scripts/generateEvmApiCloud.ts
```diff
@@ -0,0 +1,11 @@
+import { createCloudFile } from './utils/generateApiCloudCode';
+import { fetchEndpoints } from './utils/prepareSwaggerForApi';
+import path from 'path';
+
+const OUTPUT_FILE = path.join(__dirname, '../src/cloud/generated', 'evmApi.ts');
+const API_SWAGGER_URL = 'https://deep-index.moralis.io/api-docs/v2/swagger.json';
```
b4rtaz, line 6: This approach will stop work in a moment (after endpoints rename).

I think we need a place where we can store the old name of each endpoint. It could be the Endpoint interface. I think we could add a new `oldName` property.

So that means we could base on endpoint descriptors here (check the [Endpoints](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/main/packages/apiUtils/src/resolvers/Endpoints.ts) class). I suppose the problem with the `balance` won't occur in this approach.

packages/solApi/src/MoralisSolApi.ts
```diff
@@ -30,11 +30,15 @@ export class MoralisSolApi extends ApiModule {
...
 
+  private readonly getBalance = this.endpoints.createFetcher(getBalance);
+
   public readonly account = {
-    getBalance: this.endpoints.createFetcher(getBalance),
+    getBalance: this.getBalance,
     getNFTs: this.endpoints.createFetcher(getNFTs),
     getPortfolio: this.endpoints.createFetcher(getPortfolio),
     getSPL: this.endpoints.createFetcher(getSPL),
+    // Support for old naming
```
b4rtaz, line 40: `@depreciated` is missing.

ErnoW: 👍

demos/parse-server-migration/scripts/generateEvmApiCloud.ts
ErnoW: Good point. Yeah that’s why we need a better setup with proxy generation. This is now using swagger json to be consistent with the Moralis-server setup. But this is not maintainable.

This is not run automatically, so we can remove this part for now for the users until we have a better approach. (There is already a remark in the docs that generated results could differ a bit from the SDK, and that this will change in the future)

I will dive a bit in a a better way to generate these proxies.

PR #658, opened 2022-09-08
Feat/streams api

sogunshola:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description


packages/integration/mockRequests/streamApi/deleteStream.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (!value) {
+    return res(ctx.status(404));
+  }
+
+  return res(
+    ctx.status(200),
+    ctx.json({
+      chainId: value,
+    }),
+  );
```
b4rtaz, line 27: Needs a real response mock.

packages/integration/mockRequests/streamApi/getSettings.ts
```diff
@@ -0,0 +1,23 @@
...
+    return res(ctx.status(401));
+  }
+
+  const value = mockGetSettingsOutput;
+
+  if (!value) {
+    return res(ctx.status(404));
+  }
+
+  return res(ctx.status(200), ctx.json(value));
```
b4rtaz, line 22: Needs a real response mock.

packages/integration/mockRequests/streamApi/getStreams.ts
```diff
@@ -0,0 +1,25 @@
...
+  if (!value) {
+    return res(ctx.status(404));
+  }
+
+  return res(
+    ctx.status(200),
+    ctx.json({
+      total: value,
+    }),
+  );
```
b4rtaz, line 24: Needs a real response mock.

packages/integration/mockRequests/streamApi/setSettings.ts
```diff
@@ -0,0 +1,20 @@
...
+    return res(ctx.status(401));
+  }
+
+  const value = mockSetSettingsOutput;
+
+  if (!value) {
+    return res(ctx.status(404));
+  }
+
+  return res(ctx.status(200), ctx.json(value));
```
b4rtaz, line 19: Needs a real response mock.

packages/integration/mockRequests/streamApi/updateStream.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (!value) {
+    return res(ctx.status(404));
+  }
+
+  return res(
+    ctx.status(200),
+    ctx.json({
+      chainId: value,
+    }),
+  );
```
b4rtaz, line 27: Needs a real response mock.

packages/apiUtils/src/resolvers/PaginatedEndpoint.ts
```diff
@@ -13,6 +13,7 @@ export interface PaginatedResult<ApiResult> {
   page_size: number;
   cursor: string;
   result: ApiResult;
+  data: ApiResult;
```
b4rtaz, line 16: Maybe we should request the API squad to normalize this property? Here should be also `result` as everywhere.

packages/moralis/package.json
```diff
@@ -17,12 +17,12 @@
...
-    "@moralisweb3/evm-utils": "^2.3.1",
-    "@moralisweb3/sol-api": "^2.3.1",
-    "@moralisweb3/sol-utils": "^2.3.1"
+    "@moralisweb3/auth": "^2.2.0",
+    "@moralisweb3/core": "^2.2.0",
+    "@moralisweb3/api-utils": "^2.2.0",
+    "@moralisweb3/evm-api": "^2.2.0",
+    "@moralisweb3/evm-utils": "^2.2.0",
+    "@moralisweb3/sol-api": "^2.2.0",
+    "@moralisweb3/sol-utils": "^2.2.0"
```
b4rtaz, line 27: This is an unwanted change.

b4rtaz (changes requested):

packages/integration/mockRequests/streamApi/getSettings.ts
sogunshola:
```
{
  "secretKey": "top_secret",
  "region": "us-east-1"
}
```

is a real response for this endpoint

packages/apiUtils/src/resolvers/PaginatedEndpoint.ts
sogunshola: I agree, to have consistency across Moralis Apis. @ErnoW what do you think?

packages/integration/mockRequests/streamApi/getSettings.ts
b4rtaz: I see.

```
 const value = mockGetSettingsOutput;

  if (!value) {
```

This condition is always true.

packages/apiUtils/src/resolvers/PaginatedEndpoint.ts
b4rtaz: I think this is the best option. Otherwise we should modify the PaginatedEndpoint interface. The current state is wrong. We don't have `result` and `data` in the same type as the definition says.

b4rtaz (changes requested):
Still I see unnecessary conditions in tests like:

```
  const value = mockCreateStreamOutput;

  if (!value) {
    return res(ctx.status(404));
  }
```

Also, we need to clarify the `result/data` problem.

sogunshola:
> Still I see unnecessary conditions in tests like:
> 
> ```
>   const value = mockCreateStreamOutput;
> 
>   if (!value) {
>     return res(ctx.status(404));
>   }
> ```
> 
> Also, we need to clarify the `result/data` problem.

True. Removing unncessary condiitions now

b4rtaz (approved):

PR #673, opened 2022-09-15
fix: support not defined contractType for NFT types.

b4rtaz:
---
name: [support not defined contractType for NFT types.](https://github.com/MoralisWeb3/Moralis-JS-SDK/commit/1ec4c2fe3003ab6a2e10a6ee5825145fd6d698ea)
---


packages/evmUtils/src/dataTypes/EvmNft/EvmNft.ts
```diff
@@ -53,7 +53,7 @@ export class EvmNft implements MoralisDataObject {
   static parse = (data: EvmNftInput, core: MoralisCore): EvmNftData => ({
     ...data,
     chain: EvmChain.create(data.chain, core),
-    contractType: validateValidEvmContractType(data.contractType),
+    contractType: noramlizeEvmNftContractType(data.contractType),
```
ErnoW, line 56: Could we not just something like:
```
    contractType: maybe(data.contractType, validateValidEvmContractType),
```

ErnoW (changes requested):

packages/evmUtils/src/dataTypes/EvmNft/EvmNft.ts
b4rtaz: In the previous version here's also a validation. I didn't change it.
https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/main/packages/evmUtils/src/dataTypes/EvmNftContractType.ts

b4rtaz: Maybe it's not needed.

b4rtaz: I've standardized the code.

```js
contractType: maybe(data.contractType, normalizeEvmNftContractType),
```

ErnoW (approved):

PR #676, opened 2022-09-16
moralis auth firebase extension + demo.

b4rtaz:
---
name: moralis auth firebase extension + demo. POC.
---

### Checklist


.changeset/shaggy-pears-poke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@moralisweb3/client-firebase-auth': patch
+---
+
+The first release of the package.
```
ErnoW, line 5:
```suggestion
Add client package for firebase authentication.
```

.github/workflows/main.yml
```diff
@@ -111,7 +111,7 @@ jobs:
         id: cache
         with:
           path: '**/node_modules'
-          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}_C5
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}_C6
```
ErnoW, line 114: I never encountered the need for changing these, what errors do you get when you need to update the caching? we probably need a better system for this.

For now this works okay, but please update also in `docs.yml` as it shares the same cache

packages/clientFirebaseAuth/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "license": "MIT",
+  "private": false,
+  "main": "./lib/index.js",
+  "types": "./lib/index.d.ts",
+  "sideEffects": false,
+  "files": [
+    "lib/*"
+  ],
+  "scripts": {
+    "lint": "eslint . --ext .js,.ts,.tsx,jsx",
```
ErnoW, line 14: Add `format` script

packages/clientFirebaseAuth/src/requestChallenge.ts
```diff
@@ -0,0 +1,56 @@
...
+  });
+}
+
+export function requestSolanaChallenge(auth: MoralisAuth, params: RequestSolanaChallengeParams): Promise<Challenge> {
+  return requestChallenge(auth, {
+    networkType: 'solana',
+    solAddress: params.address,
+    solNetwork: params.network,
+  });
+}
```
ErnoW, line 41: Lets rename to requesMessage instead to ensure consistency with the Auth product and SDL

ErnoW, line 41: Does it make sense to have 1 function `requestMessage`, just like we have in the sdk? And let it accept a `network` param? Then we can check on the provided network (if one is provided, otherwise default to 'evm')

ErnoW (changes requested):

.github/workflows/main.yml
b4rtaz: This happens every time when I add a new package. [Details](https://github.com/MoralisWeb3/Moralis-JS-SDK/actions/runs/3083132566/jobs/4983702911).

packages/clientFirebaseAuth/package.json
b4rtaz: We don't have the `format` script anymore. For example: [auth](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/main/packages/auth/package.json), [evmApi](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/main/packages/evmApi/package.json).

b4rtaz: We have the `format` script in the root package.json only.

packages/clientFirebaseAuth/src/requestChallenge.ts
b4rtaz: I think the default `networkType` parameter is risky. Because this forces another required parameters. [I've changed](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/feat/firebase-auth-extension/packages/clientFirebaseAuth/src/requestMessage.ts) to a similar approach like in the /auth package:

```ts
RequestEvmMessageParams | RequestSolanaMessageParams
```

packages/clientFirebaseAuth/package.json
ErnoW: Ah my bad. You're right

packages/clientFirebaseAuth/src/requestChallenge.ts
ErnoW: Yes this seems perfect

demos/firebase-auth-ext/firebase.json
```diff
@@ -0,0 +1,25 @@
...
+      "yarn --cwd \"$RESOURCE_DIR\" run lint",
+      "yarn --cwd \"$RESOURCE_DIR\" run build"
+    ]
+  },
+  "emulators": {
+    "hosting": {
+      "port": 5555
+    }
+  }
+}
```
ErnoW, line 25: When I create this file I get additional properties, should those not be included?:
```
  "extensions": {
    "moralis-auth": "moralis/moralis-auth@0.1.2"
  }
```

```
  "emulators": {
    "functions": {
      "port": 5001
    },
    "hosting": {
      "port": 5002
    },
    "ui": {
      "enabled": true
    }
  },
```

ErnoW: Yeah now that I see how extensions work in the project, it makes sense to not include it in the json. Seems good.

ErnoW (approved):

PR #694, opened 2022-09-21
(feat):adding creation of EVMaddress using ens

3scava1i3r:
---
name: 'adding creation of EVMaddress using ens'
about: Currently, EvmAddress are created via EvmAddress.create("0x12423fsa"). A nice feature would be to create an address via ENS.
This will be an async call so we can add a new static method like await EvmAddress.createFromENS("name.eth")
---

### Checklist

### Solution Description

- Added ethersjs based name resolution for evmaddress
- Added tests for the same
To do
- support for user passed parameter of specific rpc urls while creating provider

### DIFF
--- a/packages/evmUtils/src/dataTypes/EvmAddress/EvmAddress.ts
+++ b/packages/evmUtils/src/dataTypes/EvmAddress/EvmAddress.ts
@@ -7,8 +7,11 @@ import {
   MoralisCoreError,
   MoralisData,
 } from '@moralisweb3/core';
+import { InputChainId } from '../EvmChain';
 import { getAddress, isAddress } from '@ethersproject/address';
 import { EvmUtilsConfig } from '../../config/EvmUtilsConfig';
+import { isValidName } from '@ethersproject/hash';
+import { getDefaultProvider } from '@ethersproject/providers';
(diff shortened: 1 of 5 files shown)


packages/evmUtils/src/dataTypes/EvmAddress/EvmAddress.ts
```diff
@@ -58,6 +61,30 @@ export class EvmAddress implements MoralisData {
...
+      throw new MoralisCoreError({
+        code: CoreErrorCode.INVALID_ARGUMENT,
+        message: 'Invalid name provided',
+      });
+    }
+
+    const provider = getDefaultProvider(network);
+    const address = await provider.resolveName(name);
+
+    return this.create(address!, core);
```
b4rtaz, line 85: I think we shouldn't mix approaches. The evmUtils package shouldn't use browser features.

Maybe this is a good point to make a new API endpoint?

b4rtaz (changes requested):

ErnoW:
Will not be implemented for now, as it we need to add extra complexity to allow for custom providers. This feature is not worth that extra complexity.

PR #699, opened 2022-09-21
Fix parse api errors

ErnoW:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

Parse API errors when api returns a 'message' in the error response

### DIFF
--- a/packages/core/src/controllers/RequestController.ts
+++ b/packages/core/src/controllers/RequestController.ts
@@ -1,4 +1,4 @@
-import axios, { AxiosError, AxiosRequestConfig } from 'axios';
+import { AxiosError, AxiosRequestConfig } from 'axios';
 import { CoreErrorCode, MoralisCoreError } from '../Error';
 import { AxiosRetry, AxiosRetryConfig } from './AxiosRetry';
 import { isTest } from '../environment/isTest';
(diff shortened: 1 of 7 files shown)


packages/core/src/controllers/RequestController.ts
```diff
@@ -1,15 +1,45 @@
...
+  if (Array.isArray(message)) {
+    return message.join(', ');
+  }
+
+  if (typeof message === 'string') {
+    return message;
+  }
+
+  return 'Unknown error (no error info returned from API)';
+};
```
b4rtaz, line 41: That part could be in the other place. New file? `AxiosError.ts`?

b4rtaz (changes requested):

packages/core/src/controllers/RequestController.ts
ErnoW: Cleaned it up to a separate file, renamed it and add description.

I'm not too happy with this approach though. As it couples api behaviour in our core package. Ideally we want this specified in the separate api packages or generally in the apiUtils package. Maybe by providing a custom errorhandler. So that could be a nice future improvement.

b4rtaz (approved):

PR #713, opened 2022-09-29
feat: add types to parse streams webhook data

ErnoW:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description


b4rtaz:
I wonder, maybe we should move streams types (EvmStream, EvmStreamResult etc) to a new package: `@moralisweb3/common-streams`. Why common? Because if I need to transfer some value from the back-end side to the front-end side we should create these types as a "common" side.

ErnoW:
> I wonder, maybe we should move streams types (EvmStream, EvmStreamResult etc) to a new package: `@moralisweb3/common-streams`. Why common? Because if I need to transfer some value from the back-end side to the front-end side we should create these types as a "common" side.

Interesting point. Do you think there will be any use-cases to use these types directly in the client? I think that this feature is mostly backend-only, and that the frontend will used partial data from this, mainly via specific endpoints/database lookups.

If there is no use-case then I'd prefer to keep it simple and contained in the streams package.

b4rtaz:
> Interesting point. Do you think there will be any use-cases to use these types directly in the client?

I'm not sure. But I can imagine that use-case.

Basically when the beta will be merged, **all SDK types** will be front-end/back-end compatible. If we don't move streams types to the common side, then we have an exception for the new approach.

ErnoW:
True I will move it for consistency sake

ErnoW:
@b4rtaz Moved everything in common-streams-utils

Also note, probably for your refactoring, you need the changes in `jest.config.js` in order to resolve tests correctly when we have packages in subfolders inside `/packages`

b4rtaz (approved):

PR #722, opened 2022-10-04
feat: add advancedOptions to streams

ErnoW:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

Add advancedOptions to streams

### DIFF
--- a/packages/integration/test/streamApi/createStream.test.ts
+++ b/packages/integration/test/streamApi/createStream.test.ts
@@ -39,6 +39,33 @@ describe('Create stream', () => {
     expect(result.result.chainIds).toContain('0x3');
   });
 
+  it('should create a stream with advanced options', async () => {
+    const result = await StreamApi.add({
+      chains: ['0x3'],
+      tag: 'test',
+      description: 'test',
+      webhookUrl: 'https://webhook.site/4f1b1b1b-1b1b-4f1b-1b1b-1b1b1b1b1b1b',
+      networkType: 'evm',
+      advancedOptions: [
+        {
+          topic0: 'SomeEvent(address,uint256)',
+          filter: { eq: ['myCoolTopic', '0x0000000000000000000000000000000000000000'] },
+          includeNativeTxs: true,
+        },
+      ],
+      abi: [],
+      allAddresses: true,
+      includeContractLogs: true,
+      includeInternalTxs: true,
+      includeNativeTxs: true,
+      topic0: ['SomeEvent(address,uint256)'],
+    });
+
+    expect(result).toBeDefined();
+    expect(result).toEqual(expect.objectContaining({}));
+    expect(result.result.chainIds).toContain('0x3');
+  });
+
   it('should not create stream', async () => {
     const failedResult = await StreamApi.add({
       chains: ['0x3'],
--- a/packages/streams/src/generated/types.ts
+++ b/packages/streams/src/generated/types.ts
@@ -134,7 +134,7 @@ export interface components {
       /** Format: double */
       gas?: number;
     };
-    IAbi: { [key: string]: components["schemas"]["AbiItem"] };
+    IAbi: { [key: string]: components["schemas"]["AbiItem"][] };
     IERC20Transfer: {
       transactionHash: string;
       contract: string;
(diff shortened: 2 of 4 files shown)


sogunshola (approved):
LGTM⚡️

packages/integration/test/streamApi/createStream.test.ts
```diff
@@ -39,6 +39,33 @@ describe('Create stream', () => {
...
+      includeInternalTxs: true,
+      includeNativeTxs: true,
+      topic0: ['SomeEvent(address,uint256)'],
+    });
+
+    expect(result).toBeDefined();
+    expect(result).toEqual(expect.objectContaining({}));
+    expect(result.result.chainIds).toContain('0x3');
+  });
+
```
b4rtaz, line 68: This test doesn't check that, the `advancedOptions` parameter is passed correctly. Check [this mock](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/main/packages/integration/mockRequests/evmApi/getTokenPrice.ts) and [these tests](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/main/packages/integration/test/evmApi/getTokenPrice.test.ts).

b4rtaz (changes requested):

packages/integration/test/streamApi/createStream.test.ts
ErnoW: Yeah that is correct. But I'm not sure if this is the correct approach.
This will not test any logic in the code. It adds extra logic in the mock, and we test that that logic works.

With mocking apis like this we can only properly test:
- Input: make sure that the input is parsed correctly (via our resolvers)
- Output: different kinds of outputs should be parsed correctly
- Errors: different kinds of errors should be handled correctly

There is no good way for us with these tests to test that certain inputs will return certain outputs. 

If we want to test that, then we need to look into other testing methods, where we call the actual API. which is something worth doing

b4rtaz: >  This will not test any logic in the code. It adds extra logic in the mock, and we test that that logic works.

I think my current implementation is too complex and the main idea is not obvious. So I've deleted all conditions for these tests and now it should be more clear. Check [this](https://github.com/MoralisWeb3/Moralis-JS-SDK/pull/723) pull request.

> There is no good way for us with these tests to test that certain inputs will return certain outputs.

IMO this claim is incorrect. Because the main idea of integration tests is that, we want to check if a specific input (or some sequence) returns a specific output. Especially for stateless REST services.

So, **we should test scenarios/cases**. And of course we can test many scenarios. Basically, we should have the same number of jest tests as we have dedicated mocks in the mock server.

This is common practice. And I can achieve it not only by a code. For example, I've created 3 test mocks for the `getTokenPrice` endpoint using the Postman. I can use the Postman mock server to create 3 integration tests.

image

Check these URLS:

```
https://ff0c1798-b3ac-4987-b9f4-58cab995a194.mock.pstmn.io/erc20/0x0000000000000000000000000000000000000001/price?chain=0x1
https://ff0c1798-b3ac-4987-b9f4-58cab995a194.mock.pstmn.io/erc20/0x0000000000000000000000000000000000000002/price?chain=0x13881
https://ff0c1798-b3ac-4987-b9f4-58cab995a194.mock.pstmn.io/erc20/0x0000000000000000000000000000000000000003/price?chain=0x1&to_block=512
```

This approach is compatible with the "[Red, Green, Refactor](https://www.codecademy.com/article/tdd-red-green-refactor)". Because when you get a bug, at the beginning you should write a "red" test. After this you should fix your code and make the test green. This is exactly that what I've done in [this pull](https://github.com/MoralisWeb3/Moralis-JS-SDK/pull/720) request.

After all, I think the main reason why we need tests/integration tests is that, we want to have a proof that the code works properly. And my remarks refer to that, I don't see the proof that, the `advancedOptions` parameter is passed to the backend service.

ErnoW: Ok I see now what you mean. Yeah this way we can check if the params are correctly provided via check on the server. We should keep the logic minimal though. I like the idea with 'scenarios' the most. I did a similar approach with some other tests.

We can generalise it to specify scenarios like:
```
{
  condition: // all the checks that we verify for, most likely these are url-params, and request body values
  response: // returned response data
  responseStatus: // returned response status
}
```

b4rtaz (approved):

packages/integration/test/streamApi/createStream.test.ts
b4rtaz: Yes, exactly. Maybe there is some package to handle it nicely. But for now the `lodash` approach is fine too.

PR #741, opened 2022-10-11
Feat/express proxy cli

sogunshola:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

### DIFF
--- a/packages/apiUtils/src/resolvers/Endpoint.ts
+++ b/packages/apiUtils/src/resolvers/Endpoint.ts
@@ -7,8 +7,22 @@ export enum EndpointBodyType {
   BODY = 'body',
 }
 
+export type EndpointGroup =
+  | 'block'
+  | 'events'
+  | 'nft'
+  | 'token'
+  | 'transaction'
+  | 'utils'
+  | 'balance'
+  | 'resolve'
+  | 'defi'
+  | 'ipfs'
+  | 'account';
+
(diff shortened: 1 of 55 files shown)


packages/apiUtils/src/resolvers/Endpoint.ts
```diff
@@ -7,8 +7,22 @@ export enum EndpointBodyType {
...
+  | 'utils'
+  | 'balance'
+  | 'resolve'
+  | 'defi'
+  | 'ipfs'
+  | 'account';
+
 export interface Endpoint<ApiParams, Params, ApiResult, AdaptedResult, JSONResult> {
   name: string;
+  group?: EndpointGroup;
```
b4rtaz, line 25: I think this approach is wrong. Now we have two places where I can define a "group":

* the `MoralisEvmApi` class
* the `group` property in the `Endpoint` interface.

So that can generate conflicts. I think we should keep the [single source of truth](https://en.wikipedia.org/wiki/Single_source_of_truth) principle. So we should have that information in only one place. Basically we already have that place (the MoralisEvmApi class). So we should figure out how to access to this information without a duplication.

b4rtaz (changes requested):

ErnoW:
Closing, as this will be fixed with the expressJs package and the new way we handle operations

PR #742, opened 2022-10-11
feat: export utils in umbrella package

sogunshola:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description


b4rtaz:
I still think we should not do it in this way.

@ErnoW ^

ErnoW:
> I still think we should not do it in this way.
> 
> @ErnoW ^

What is your main concern @b4rtaz ?
I still think this offers a better dev-experience, as most people are only concerned in using the umbrella package. This way they don't have to worry about importing types from one of the specific packages.

We could change the approach a bit to export only certain functions/classes instead of everything. For example
```
export {
    EvmAddress,
} from "@moralisweb3/evm-utils";
```
(instead of export * from `"@moralisweb3/evm-utils"`).

sogunshola:
Are there types we don't wish to expose from the umbrella package?

b4rtaz:
> What is your main concern @b4rtaz ?

@ErnoW I think the biggest downside here is that, users can choose different imports in different files. Sometimes it would be:

```ts
import { EvmAddress } from "moralis";
```

Sometimes:

```ts
import { EvmAddress } from "@moralisweb3/evm-utils";
```

This leads to a mess in a project. By allowing this we allow to make bad practices.

ErnoW:
> > What is your main concern @b4rtaz ?
> 
> @ErnoW I think the biggest downside here is that, users can choose different imports for different files. Sometimes it would be:
> 
> ```ts
> import { EvmAddress } from "moralis";
> ```
> 
> Sometimes:
> 
> ```ts
> import { EvmAddress } from "@moralisweb3/evm-utils";
> ```
> 
> This leads to a mess in a project. By allowing this we allow to make bad practices.

I see what you mean. But that does not need to be too bad. I think the benefits outweighs this downside. If we export all data-types/types/utils that people most likely are using from the umbrella package, then that means a better dev-experience (especially for new users). The recommended import will be from the umbrella package for most use-cases.

In the meantime, we still have a modular structure, with isolated logics, and the option to only install certain packages.

And since `@moralisweb3/evm-utils` and `moralis` have versioning locked together, will mean that
```
import { EvmAddress } from "@moralisweb3/evm-utils";
```
and 
```
import { EvmAddress } from "moralis";
```
Will result in the exact same import.

In the docs etc. we can only show  `import { EvmAddress } from "moralis"` (as it is easier to implement and onboard users). But in the technical docs we can also show the `import { EvmAddress } from "@moralisweb3/evm-utils"` import.

The imports for specific packages will probably only be used by power-users or people making libraries/utilities on top of the moralis SDK.

sogunshola:
PR updated with named exports for `core` and utility packages. Not so sure if the core is so important since we have that as `Moralis.core` but I added it anyways so now users can do something like `import { isMoralisError } from 'moralis'`

ErnoW:
What about this?:
```

export * from '@moralisweb3/evm-utils';
export * from '@moralisweb3/sol-utils';
export {
  Types as StreamTypes,
} from '@moralisweb3/streams';
export * from '@moralisweb3/streams/lib/dataTypes';
export {
  ApiErrorCode,
  AuthErrorCode,
  BigNumber,
  BigNumberPrimitive,
  BigNumberish,
  CoreErrorCode,
  EvmAddressFormat,
  EvmChainIdFormat,
  LogLevel,
  MoralisApiError,
  MoralisAuthError,
  MoralisCoreError,
  MoralisCoreProvider,
  MoralisError,
  MoralisErrorCode,
  MoralisErrorDetails,
  MoralisErrorOptions,
  MoralisStreamError,
  StreamErrorCode,
  isMoralisError,
} from '@moralisweb3/core';
```

- So we export anything from our util packages, as these serve a lot of use-cases
- We skip exporting api-utils, as it is used internally
- We export the `StreamTypes` under a better name
- We export only things from core that are really valuable for clients, and skip the things that we are mostly using internally

We might want to export request/param types from evmApi/streas/auth etc. But lets consider that after the refactoring

b4rtaz:
> What about this?:

I think this increases unnecessarily a complexity. Now I need to remember to add new types to this export list. If I forget it, then we have a no consistency and a mess on the end-user side. I think the `export * from 'X'` approach decreases this problem to 0.

Maybe there is a third solution. The Firebase uses it. Check [this repo](https://github.com/firebase/firebase-js-sdk/tree/master/packages/firebase) ([app](https://github.com/firebase/firebase-js-sdk/tree/master/packages/firebase/app), [functions](https://github.com/firebase/firebase-js-sdk/tree/master/packages/firebase/functions), ...). Basically we should create "virtual folders" in the `moralis` package. The final result would look like this:

```ts
import { EvmAddress } from 'moralis/evm-utils';
import { MoralisError } from 'moralis/core';
```

We have what we want - every types in the umbrella package. Additionally this approach prevents naming conflicts.

b4rtaz:
Some Firebase example:

```ts
import { httpsCallable } from '@firebase/functions';
import { httpsCallable  } from 'firebase/functions';

// `firebase/functions` is a map to `@firebase/functions`
```

ErnoW:
Changing it to imports from subpaths could be a nice-to-have. And this is less confusing for users as it is still the same package. I like that approach. 👍 

I still believe that there is a case to be made to export classes (mainly the dataTypes) directly from the umbrella package. This will allow us to have more 'magic' one-line of codes when onboarding users. For example in our api docs we now have:
```typescript
import Moralis  from 'moralis';
import { EvmChain } from '@moralisweb3/evm-utils';

const chain = EvmChain.ETHEREUM;

await Moralis.start({
    apiKey: '<YOUR_API_KEY>',
    // ...and any other configuration
});
```

Or if someone wants to use a BigNumber, EvmChain their code will look like:
```typescript
import Moralis  from 'moralis';
import { EvmChain } from '@moralisweb3/evm-utils';
import { BigNumber } from '@moralisweb3/core';
```

For new users this adds extra complexity
- Extra lines of code
- Confusing about where to import from (bignumber is in core, evmChain is in evm-utils)
- Confusing if needs to install extra packages (this is solved though with the subpath approach)

These kind of snippets are really commonly used and need as good as a dev-experience as possible. Importing EvmAddress from moralis directly solves this issues.

There might not be a use-case for other classes/methods/types though.
But especially some datatypes like EvmAddress, Bignumber etc. have use-cases

b4rtaz:
> For new users this adds extra complexity

I agree. I think the sub-paths approach reduces my concerns. Now we have a simple mapping: `moralis/x` -> `@moralisweb3/x`. This is reasonable and easy to maintain.

sogunshola:
I agree. I also think the sub-path approach is nice, but also having it exported directly from the umbrella package is also nice.

I'll push a solution with both implementations, having both can't hurt 🙂

sogunshola:
New push, now imports can also be done this way

```typescript
import {} from 'moralis/common-evm-utils';
import {} from 'moralis/common-sol-utils';
import {} from 'moralis/core';
```

as well as `import {} from 'moralis';`

b4rtaz:
@sogunshola I think it would be better if we have a single approach (like the Firebase). Now an end user can be confused, which path is correct:

```
import { EvmAddress } from 'moralis/common-evm-utils'; // 1
import { EvmAddress } from 'moralis'; // 2
import { EvmAddress } from '@moralisweb3/common-evm-utils'; //3
```

I would reduce it to two options (`1` and `3`).

packages/moralis/package.json
```diff
@@ -16,6 +16,32 @@
...
+      ],
+      "core": [
+        "lib/core/index.d.ts"
+      ],
+      "common-evm-utils": [
+        "lib/evmUtils/index.d.ts"
+      ],
+      "common-sol-utils": [
+        "lib/solUtils/index.d.ts"
+      ],
```
b4rtaz, line 39: Should be:

```
     "common-evm-utils": [
        "lib/commonEvmUtils/index.d.ts"
      ],
      "common-sol-utils": [
        "lib/commonSolUtils/index.d.ts"
      ],
```

b4rtaz (changes requested):

b4rtaz (approved):

PR #773, opened 2022-10-24
feat: Data tracking on Product usage

sogunshola:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

### DIFF
--- a/packages/core/src/Config/CoreConfig.ts
+++ b/packages/core/src/Config/CoreConfig.ts
@@ -22,4 +22,9 @@ export const CoreConfig = {
     name: 'defaultNetwork',
     defaultValue: 'Evm',
   } as ConfigKey<Network>,
+
+  product: {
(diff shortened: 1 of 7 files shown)


packages/core/src/Config/CoreConfig.ts
```diff
@@ -22,4 +22,9 @@ export const CoreConfig = {
     name: 'defaultNetwork',
     defaultValue: 'Evm',
   } as ConfigKey<Network>,
+
+  product: {
+    name: 'product',
+    defaultValue: 'JS SDK',
```
b4rtaz, line 28: I think the default value should be null/undefined. If there is no product, we shouldn't send that header.

b4rtaz (changes requested):

b4rtaz (approved):

PR #775, opened 2022-10-25
serialization tests for operations.

b4rtaz:
---
name: serialization tests for operations.
---


packages/common/evmUtils/src/operations/token/getWalletTokenTransfersOperation.ts
```diff
@@ -44,7 +44,7 @@ export const getWalletTokenTransfersOperation: PaginatedOperation<
 
 function getRequestUrlParams(request: GetWalletTokenTransfersRequest, core: MoralisCore) {
   return {
-    address: String(request.address),
+    address: EvmAddress.create(request.address, core).checksum,
```
ErnoW, line 47: Should be lowercase. Most of our APIs used to only work with lowercase. Not entirely sure if it is still the case everywhere but lets always cast to lowercase with api requests.

ErnoW (changes requested):

packages/common/evmUtils/src/operations/token/getWalletTokenTransfersOperation.ts
b4rtaz: Good point! I've changed it.

sogunshola (approved):

ErnoW (approved):

PR #787, opened 2022-11-01
Endpoints and types refactor for evm

ErnoW:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description


packages/common/evmUtils/src/operations/nft/getNFTContractMetadataOperation.ts
```diff
@@ -0,0 +1,83 @@
...
+
+type PathParams = operations[OperationId]['parameters']['path'];
+type QueryParams = operations[OperationId]['parameters']['query'];
+type RequestParams = PathParams & QueryParams;
+
+type SuccessResponse = operations[OperationId]['responses']['200']['content']['application/json'];
+
+// Exports
+
+export type GetNftContractMetadataRequest = Camelize<Omit<RequestParams, 'chain' | 'address'>> & {
```
b4rtaz, line 16: `GetNftContractMetadataRequest` -> `GetNFTContractMetadataRequest`

b4rtaz (changes requested):
@sogunshola  *Request & *Response types should have same prefix as OperationId (same uppercase lowercase letters). You may run the `yarn gen:client` command for the `@moralisweb3/client-evm-api` and the `@moralisweb3/client-sol-api` packages and check that, the generated clients have correct references.

packages/common/evmUtils/src/operations/defi/getPairReservesOperation.ts
```diff
@@ -0,0 +1,82 @@
...
+  deserializeRequest,
+  deserializeResponse,
+};
+
+// Methods
+
+function getRequestUrlParams(request: GetPairReservesRequest, core: Core) {
+  return {
+    chain: EvmChainResolver.resolve(request.chain, core).apiHex,
+    pairAddress: EvmAddress.create(request.pairAddress, core).lowercase,
```
Y0moo, line 53: Shouldn't it be `pair_address` as we have in openapi?

Y0moo: I can see some renaming to camelcase across getRequestUrlParams functions. But as I understand params inside should be equal to `query`+`path`?

Y0moo: @sogunshola

b4rtaz: Good point! The `getRequestUrlParams` should return names using by the API (especially search parameters, url parameters depend on the `urlPathPattern` property).

Y0moo: I guess we simply can declare in function returned type `RequestParams`:
```
function getRequestUrlParams(request: GetPairAddressRequest, core: Core) : RequestParams { }
```

Y0moo: Maybe make a type `Record<keyof RequestParams, string | undefined>`, but I don't like this approach since it makes code more complicated. I guess it'd be better to put this check to the tests

sogunshola: There is a small problem here.
```typescript
public prepareUrl(baseUrl: string, request: Request) {
    const urlParams = this.operation.getRequestUrlParams(request, this.core);

    let urlPath = this.operation.urlPathPattern;

    for (const paramName of this.operation.urlPathParamNames) {
      const paramValue = urlParams[paramName as string];
      if (!paramValue) {
        throw new Error(`Param ${paramName as string} is required`);
      }
      urlPath = urlPath.replace(`{${paramName as string}}`, paramValue as string);
    }
    const url = `${baseUrl}${urlPath}`;

    const urlSearchParams: Record<string, string | string[]> = {};
    Object.keys(urlParams)
      .filter((paramName) => !this.operation.urlPathParamNames.includes(paramName as keyof Request))
      .forEach((paramName) => {
        const paramValue = urlParams[paramName];
        if (paramValue) {
          urlSearchParams[paramName] = paramValue;
        }
      });

    return { url, urlSearchParams };
  }
```

in this logic, the `urlPathParamNames` is `(keyof Request)[]` which has been camelized so the `paramName` searched for in the `urlParams` has to be camelCased as well hence the change to camel case. this is not the case for the search parameters because this logic does not make use of `urlSearchParamNames`. To change this to snake case, we have to change the expected type for `urlPathParamNames`.
@b4rtaz @Y0moo

packages/common/evmUtils/src/operations/operations.test.ts
```diff
@@ -14,11 +14,25 @@ describe('operations', () => {
...
+        const ignoreBodyCheckOperationNames = ['abi'];
+        const openApiBodyParamNames = reader
+          .readOperationRequestBodyParamNames(operation.id)
+          ?.map(toCamel)
+          .filter((name) => !ignoreBodyCheckOperationNames.includes(name));
 
-        expect(operation.urlPathParamNames.sort().join(',')).toBe(openApiPathParamNames?.sort().join(','));
+        ignoreBodyCheckOperationNames.forEach((name) => {
+          operation.bodyParamNames?.splice(operation.bodyParamNames.indexOf(name as never), 1);
+        });
```
b4rtaz, line 25: I think this is too complicated. You can simplify this:

```ts
 const ignoreBodyCheckOperationNames = ['getContractEvents', 'uploadFolder'];
// ...

if (!ignoreBodyCheckOperationNames.includes(operation.name)) {
        expect(bodyParamNames?.sort().join(',')).toBe(openApiBodyParamNames?.sort().join(','));
}
```

b4rtaz (changes requested):

packages/common/evmUtils/src/operations/operations.test.ts
b4rtaz: We want skip *operations*, not *parameter names*.

b4rtaz (approved):

PR #789, opened 2022-11-02
auto generated front-end api client.

b4rtaz:
---
name: auto generated front-end api client.
---


ErnoW:
Looks good 👍 
One thing that comes to mind is that we might want to separate generated files more clearly from non-generated files. Previously I put these files in a `/generated` folder. This makes it more obvious that we should not edit these files, and it is easy to ignore these files for prettier via a glob pattern.

What do you think about that? The downside is that the file structure becomes less intuitive if we have lots of generated files though.

b4rtaz:
@ErnoW I think this is a good practice, especially at the start when a developer is not a familiar with the code base. Updated it.

ErnoW (approved):

PR #792, opened 2022-11-07
separate common-core and umbrella package configs

FedericoAmura:
---
name: 'common-core parameters not using all possible options from umbrella'
about: A new pull request
---

### Checklist

### Issue Description

This PR solves that when starting the common core package you are offered to pass an apiKey, which is not needed and should not be used on FE

### DIFF
--- a/packages/common/core/src/Config/MoralisConfig.ts
+++ b/packages/common/core/src/Config/MoralisConfig.ts
@@ -4,33 +4,32 @@ import { EvmChainish } from './interfaces/EvmChainish';
 
 // @moralisweb3/common-core
 type CoreConfigType = typeof CoreConfig;
-type CoreConfigValues = Omit<{ [Key in keyof CoreConfigType]: CoreConfigType[Key]['defaultValue'] }, 'product'>;
+export type CoreConfigValues = Omit<{ [Key in keyof CoreConfigType]: CoreConfigType[Key]['defaultValue'] }, 'product'>;
 
 // @moralisweb3/common-evm-utils
-interface EvmUtilsConfigValues {
+export interface EvmUtilsConfigValues {
   formatEvmAddress: EvmAddressFormat;
   formatEvmChainId: EvmChainIdFormat;
 }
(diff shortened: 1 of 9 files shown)


packages/moralis/src/Config/MoralisConfig.ts
```diff
@@ -0,0 +1,3 @@
+import { ApiUtilsConfigValues, MoralisCoreConfigValues } from '@moralisweb3/common-core/src/index';
```
sogunshola, line 1:
```suggestion
import { ApiUtilsConfigValues, MoralisCoreConfigValues } from '@moralisweb3/common-core';
```

packages/common/core/src/Config/MoralisConfig.ts
```diff
@@ -4,33 +4,32 @@ import { EvmChainish } from './interfaces/EvmChainish';
...
+export interface EvmUtilsConfigValues {
   formatEvmAddress: EvmAddressFormat;
   formatEvmChainId: EvmChainIdFormat;
 }
 
 // @moralisweb3/api-utils
-interface ApiUtilsConfigValues {
+export interface ApiUtilsConfigValues {
   apiKey: string;
 }
```
b4rtaz, line 19: Maybe this part should be moved to the `@moralisweb3/api-utils` package? If this is interface is not used in the `core` package, it shouldn't be there.

b4rtaz (changes requested):

packages/moralis/src/Config/MoralisConfig.ts
FedericoAmura: Fixed

packages/common/core/src/Config/MoralisConfig.ts
FedericoAmura: You are right, missed it
Fixed

b4rtaz (approved):

sogunshola (approved):

PR #816, opened 2022-11-15
Feat/streams refactor

b4rtaz:
---
name: Streams refactor
---


packages/streams/src/operations/evmStreams/addAddressEvmOperation.ts
```diff
@@ -0,0 +1,95 @@
...
+  } else {
+    address = EvmAddress.create(request.address, core).checksum;
+  }
+  return {
+    address,
+  };
+}
+
+function deserializeResponse(jsonResponse: AddAddressEvmJSONResponse) {
+  return jsonResponse;
```
sogunshola, line 68: We should return address as `EvmAdress` here and also check if it is an array or not.

sogunshola (changes requested):

packages/streams/src/operations/evmStreams/addAddressEvmOperation.ts
b4rtaz: Good point! I've just fixed it.

PR #824, opened 2022-11-16
@moralisweb3/next

Y0moo:
---
name: 'Pull request'
about: Auto-generated `@moralisweb3/next`
---

### Checklist

### Issue Description

### Solution Description

All hooks are generated automatically except ''auth" hooks. Readme is generated automatically but requires some manual editing because of unexpected 'plopjs' behavior (cuts text for Append action). 

Updated `demos/nextjs`


Y0moo:
@ErnoW `yarn generate` => `next_readme` to get buggy Handlebard behavior. It cuts 'append' action to "Response" for `useEvmPairAddress`

Expected readme:

```
### Response:

{
  token0: {
    token: Erc20Token;
    blockNumber: string | undefined;
    validated: number | undefined;
    createdAt: Date | undefined;
  };
  token1: {
    token: Erc20Token;
    blockNumber: string | undefined;
    validated: number | undefined;
    createdAt: Date | undefined;
  };
  pairAddress: EvmAddress | undefined;
};
```
Actual result:

```
### Response:
{
  token0: {
    token: Erc20Token;
    blockNumber: string | undefined;
    validated: number | undefined;
    createdAt: Date | undefined;
  };
  token1: {
    token: Erc20Token;
    blockNumber: string | undefined;
    validated: number | undefined;
    createdAt: Date | undefined;
  };
  pairAddress: EvmAddress |
```

Y0moo:
Deps check will be fixed after https://github.com/MoralisWeb3/Moralis-JS-SDK/pull/825 is merged

packages/codegen/src/next/generators/hooks/templates/hook.ts.hbs
```diff
@@ -0,0 +1,23 @@
...
+import { 
+  {{ names.operation }} as operation, 
+  {{ names.request }}, 
+  {{ names.response }} 
+} from '{{ names.commonUtils }}';
+import { SWRConfiguration } from 'swr/dist/types';
+import useSWR from 'swr';
+
+export const {{ names.hook }} = (request: {{ names.request }}, SWRConfig?: SWRConfiguration) => {
+  const { data, error, mutate, isValidating } = useSWR<{{ names.response }}>(
```
b4rtaz, line 11: I would add here:

```
// CAUTION: This file is automatically generated. Do not edit it manually!
```

packages/codegen/src/utils/ModuleGenerator.ts
```diff
@@ -0,0 +1,63 @@
...
+  public get operationsPackageName() {
+    switch (this.module) {
+      case 'evmApi':
+        return 'moralis/common-evm-utils';
+      case 'solApi':
+        return 'moralis/common-sol-utils';
+      case 'auth':
+        return '@moralisweb3/auth';
+      default:
+        throw new Error('Not implemented');
```
b4rtaz, line 21:
```ts
throw new Error(`Not implemented ${this.module}`);
```

 This approach increases errors readability.

packages/codegen/src/utils/OperationFilesParser.ts
```diff
@@ -0,0 +1,51 @@
+import _ from 'lodash';
+import { ModuleGenerator } from './ModuleGenerator';
+import { Project, TypeFormatFlags } from 'ts-morph';
+import path from 'node:path';
+import prettier from 'prettier';
+
+export class OperationFilesParser extends ModuleGenerator {
```
b4rtaz, line 7: Inheritance is passé. [Composition is better](https://betterprogramming.pub/inheritance-vs-composition-2fa0cdd2f939).

packages/common/core/src/Core.ts
```diff
@@ -28,7 +28,11 @@ export class Core {
   }
 
   public readonly name = Core.moduleName;
-  private isStarted = false;
+  private _isStarted = false;
+
+  public get isStarted() {
+    return this._isStarted;
+  }
```
b4rtaz, line 35: I think we don't need it. The above layer should know how that, how many times executed the start method.

packages/next/src/MoralisNextApi/RequestHandlerResolver.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (!module) {
+      return null;
+    }
+
+    const operation = module.operations.find((op) => op.name === operationName);
+    if (!operation) {
+      return null;
+    }
+
+    return new OperationResolver(operation, module.baseUrl, Moralis.Core);
```
b4rtaz, line 48: I wonder, how an usage with a pagination look like?

b4rtaz (changes requested):

ErnoW (changes requested):
Very nice work! 🤩
See my comments for some feedback

packages/common/core/src/Core.ts
b4rtaz: I agree that, it can be valuable for beginners. But basically if you don't know that, if you initialized something or don't, that means you probable have a mess in you project.

packages/codegen/src/next/generators/hooks/templates/hook.ts.hbs
Y0moo: As I remember we have an agreement that we put all auto-generated files under './generated' folder. So I'm sure if it makes sense to add to each hook

demos/nextjs/app/components/modules/Authentication/Authentication.tsx
```diff
@@ -52,14 +54,24 @@ const Authentication = () => {
...
     const { account, chain } = await connectAsync({ connector });
 
-    const userData = { address: account, chain: chain.id, networkType: 'evm' };
+    try {
+      const challengeParams = {
+        address: account,
+        chainId: chain.id,
+        domain: 'amazing.dapp',
+        uri: 'http://localhost:3000',
+        timeout: 120,
```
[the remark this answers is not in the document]
Y0moo: What do you think about splitting the interface in the operation like this?:
```
export interface RequestChallengeEvmRequestClient {
  address: EvmAddressish;
  chainId: EvmChainish;
}

export interface RequestChallengeEvmRequestServer extends Camelize<Omit<RequestParams, 'address' | 'chainId' | 'expirationTime' | 'notBefore'>> {
  expirationTime?: DateInput;
  notBefore?: DateInput;
}

export interface RequestChallengeEvmRequest extends RequestChallengeEvmRequestClient, RequestChallengeEvmRequestServer {}
```

packages/codegen/src/next/generators/hooks/templates/hook.ts.hbs
b4rtaz: Maybe you right. But still I see scenarios when someone opens a hook by "Go to definition" and modifies that file. Of course I can be too hypersensitive for that problem.  😅

packages/next/src/MoralisNextApi/RequestHandlerResolver.ts
Y0moo: No pagination in this version. Do you mean any possible buggy behavior?

demos/nextjs/app/components/modules/Authentication/Authentication.tsx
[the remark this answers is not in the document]
b4rtaz: Maybe that part shouldn't be auto-generated. If the client side interface != the backend side interface, and the back-end needs an extra logic to validate an input, that could means, we shouldn't generate it.

Basically that approach is set for the [Firebase auth](https://github.com/MoralisWeb3/firebase-extensions/blob/main/auth/functions/src/index.ts). The backend here also requires an additional logic like creating a token.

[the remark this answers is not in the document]
Y0moo: What do you mean by "that part"? The authentication hooks are not auto-generated. My question is about splitting an interface into two parts so devs can easily see which params are "client" and "server"

packages/next/src/MoralisNextApi/RequestHandlerResolver.ts
b4rtaz: I wonder what the end-user needs to do, to make a request for a next page. How the code looks like.

```ts
const [cursor, setCursor] = useState(() => null);
const transaction = useEvmTransaction({
    cursor
});

// ...

function onNextClicked() {
    setCursor(...);
}
```

Something like this?

demos/nextjs/app/components/modules/Authentication/Authentication.tsx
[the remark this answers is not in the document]
b4rtaz: I see. But why you need to use a common type if the hook is not auto-generated? You can use own type dedicated for your layer.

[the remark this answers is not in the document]
b4rtaz: I think the operation approach  is dedicated **ONLY** for one case where the front-end client has the same parameters as the backend proxy. I wouldn't add there a next responsibility.

So that means you cannot use the operation approach for the auth.

For the Firebase you have: 

```[front-end-client] ---(client-backend-interface)---> [backend] ---(operations-interface)---> [operations]```

So the Firebase client side package has own interface to send a data to the backend. That doesn't damages assumptions of the operations approach.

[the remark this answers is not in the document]
Comment (Y0moo) [reply demos/nextjs/app/components/modules/Authentication/Authentication.tsx:63]: I see what you mean. I don't think it's the only time we will face this problem. I proposed a simple way to save us from using the `Omit` generic each time to split the operation's parameters for the client and server.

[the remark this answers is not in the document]
Comment (Y0moo) [reply demos/nextjs/app/components/modules/Authentication/Authentication.tsx:63]: I'm just aware that our authentication types can become outdated if we are not referring to `@moralisweb3/auth` in client-side packages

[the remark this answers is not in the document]
Comment (b4rtaz) [reply demos/nextjs/app/components/modules/Authentication/Authentication.tsx:63]: I see. You try to apply [the DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle. But "[duplication is far cheaper than the wrong abstraction](https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction)". Here I think we have a wrong abstraction. Yes, those interfaces are similar (or the first interface is a subset of second one), but this is not a reason why why need to add a new responsibility for the operation approach. Now this approach is super simple and I would keep it.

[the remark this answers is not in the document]
Comment (Y0moo) [reply demos/nextjs/app/components/modules/Authentication/Authentication.tsx]: @ErnoW now these params are set on the server-side in `pages\api\moralis\[...moralis].ts`:
```
import { MoralisNextApi } from '@moralisweb3/next';

export default MoralisNextApi({
  apiKey: process.env.MORALIS_API_KEY,
  authentication: {
    domain: 'amazing.dapp',
    uri: process.env.NEXTAUTH_URL,
    timeout: 120,
  },
});
```

File: packages/next/src/MoralisNextApi/MoralisNextApi.ts [hunk 7]
```diff
@@ -0,0 +1,35 @@
+import { ApiUtilsConfigValues } from '@moralisweb3/api-utils';
+import { MoralisCoreConfigValues } from 'moralis/common-core';
+import { MoralisNextHandlerParams } from './types';
+import { RequestHandlerResolver } from './RequestHandlerResolver';
+import Moralis from 'moralis';
+import type { NextApiRequest, NextApiResponse } from 'next';
+
+export type MoralisConfigValues = MoralisCoreConfigValues & ApiUtilsConfigValues;
+
+async function MoralisNextHandler({ req, res }: MoralisNextHandlerParams) {
```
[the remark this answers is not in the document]
Y0moo: I noticed that Next-Auth used this syntax for defining auth resolver and auth providers. 
Example from the docs:

```
/pages/api/auth/[...nextauth].js

import NextAuth from 'next-auth'
import AppleProvider from 'next-auth/providers/apple'
import FacebookProvider from 'next-auth/providers/facebook'
import GoogleProvider from 'next-auth/providers/google'
import EmailProvider from 'next-auth/providers/email'

export default NextAuth({
  providers: [
    // OAuth authentication providers...
    AppleProvider({
      clientId: process.env.APPLE_ID,
      clientSecret: process.env.APPLE_SECRET
    }),
    FacebookProvider({
      clientId: process.env.FACEBOOK_ID,
      clientSecret: process.env.FACEBOOK_SECRET
    }),
    GoogleProvider({
      clientId: process.env.GOOGLE_ID,
      clientSecret: process.env.GOOGLE_SECRET
    }),
    // Passwordless / email sign in
    EmailProvider({
      server: process.env.MAIL_SERVER,
      from: 'NextAuth.js <no-reply@example.com>'
    }),
  ]
})
```

I didn't find any name conventions for this case. 
We shouldn't lowercase the `MoralisNextAuthProvider` so it follows 'next-auth' namings, but we can lowercase `MoralisNextApi`.

packages/next/src/MoralisNextApi/RequestHandlerResolver.ts
Y0moo:
```
const [cursor, setCursor] = useState(0);
```
Yes, this is the best way to make this atm. Also, I think we should overwrite the `string` types in the SDK to `number` in future.

Y0moo: because atm simple cursor + 1 will not work that simple, you need to make string number => calculate => string again

Y0moo: In case we are agree to make pagination options `number` type, we can add simpler way for paginating in @moralisweb3/next

ErnoW (approved):

PR #825, opened 2022-11-16
Extend functionality for findMissingDependencies script

Y0moo:
---
name: 'Pull request'
about: Extended functionality for findMissingDependencies.js
---

### Checklist

### Issue Description

The script doesn't recognize relative imports from packages(examples: `'parse/node'`, `next-auth/providers/credentials`)

### Solution Description

1. Added `isValidPackage` function
2. Removed `IGNORE_DEPENDENCIES` const since there is no need for it.


b4rtaz (changes requested):
Still this is not a final solution. This problem refers to all packages with many "exports" in the package.json file. We have this case [here](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/main/packages/moralis/package.json). The Firebase has it too:

```ts
import { Database } from 'firebase/database';
```

I understand you have updated this validator because you recognized more cases like this. I think it would be better if we don't ignore these cases, but validate it too. So I think we should map the name, like: `parse/node => parse` and check the existence of the `parse` in the package.json.

b4rtaz:
> Good point @b4rtaz. Updated the script

The `validateImports` method has a wrong name. It doesn't validate anything. Should be `readDependencies` or similar.

Y0moo:
Please check new namings @b4rtaz

b4rtaz (approved):

b4rtaz:
@Y0moo approved!

PR #826, opened 2022-11-17
chore(evmapi): change to scenario based tests

sogunshola:
---
name: EVMAPI test
about: covert transactions and events test to scenario based tests
---

### Checklist

### Issue Description


b4rtaz (dismissed):
The build crashes. Furthermore it's fine.

sogunshola:
> The build crashes. Furthermore it's fine.

@b4rtaz I split the test in `setup.ts` so test can run properly. We will remove this when all tests have been converted.

Only issue now is that Eunice and Federico have to sync with this as well as it affects them. What do you think? any better approach?

b4rtaz:
> Only issue now is that Eunice and Federico have to sync with this as well as it affects them. What do you think? any better approach?

I don't see a better option. We need to sync with these changes.

b4rtaz:
@sogunshola You may notify them.

PR #881, opened 2022-12-07
fix: Missing wei type in format type

trietsipher:
---
Missing wei type in format type)
---

Hi guys, I used the API (https://docs.moralis.io/reference/getnfttrades) but the response data in the field price is different between API and SDK because it misses the 'wei' like other files. 
Could you guys help me double-check it, thank you so much


trietsipher:
@b4rtaz My pleasure, and thanks for letting me know 😊

PR #893, opened 2022-12-12
Backend adapters - proof of concept

b4rtaz:
---
name: Backend adapters - proof of concept
---


packages/client/evmAuth/src/MetaMaskEvmConnector.ts
```diff
@@ -0,0 +1,47 @@
+import { Web3Provider } from '@ethersproject/providers';
+import { AuthClientError, AuthClientErrorCode, WalletDetails } from '@moralisweb3/client-auth-utils';
+import { EvmConnection, EvmConnector } from './EvmConnector';
+import detectEthereumProvider from '@metamask/detect-provider';
+
+export class MetaMaskEvmConnector implements EvmConnector {
+  public readonly name = 'default';
```
ErnoW, line 7: Let's rename this to metamask

demos/webpack/src/main.ts
```diff
@@ -0,0 +1,131 @@
+import { MoralisClient } from '@moralisweb3/client';
+import { FrontEndOnlyBackendAdapter } from '@moralisweb3/client-backend-adapter-frontend-only';
+import { Connection, clusterApiUrl, PublicKey } from '@solana/web3.js';
+import { WalletConnectEvmConnector } from '@moralisweb3/client-connector-wallet-connect';
+
+async function reloadCurrentUser() {
+  const userAddress = document.getElementById('userAddress') as HTMLElement;
+  const userBalance = document.getElementById('userBalance') as HTMLElement;
+
+  const users = await Promise.all([MoralisClient.EvmAuth.tryGetUser(), MoralisClient.SolAuth.tryGetUser()]);
```
ErnoW, line 10: Probably a good utility in the future is to have a `tryGetUser` on `MoralisClient`, that calls this function for all network types, like you do here

packages/common/core/src/Modules/Module.ts
```diff
@@ -1,52 +1,23 @@
...
-  public abstract start(): void | Promise<void>;
-
-  /**
-   * Any cleanup that needs to be done for removing this module.
-   * It also should remove the module via `this.core.modules.remove(this.name)`
-   */
-  public cleanUp() {
-    this.core.modules.remove(this.name);
-  }
+export interface Module {
```
ErnoW, line 13: Nice cleanup, as we don't need the eventlisteners by default anymore 🎉

packages/client/authUtils/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "lib/*"
+  ],
+  "scripts": {
+    "test": "yarn jest --runInBand --detectOpenHandles --forceExit --ci",
+    "test:watch": "yarn run test --watch",
+    "lint": "eslint . --ext .js,.ts,.tsx,jsx",
+    "clean": "rm -rf lib && rm -rf tsconfig.tsbuildinfo && rm -rf ./node_modules/.cache/nx",
+    "build": "tsc",
+    "dev": "tsc --watch",
+    "gen:client": "node ../apiUtils/scripts/generate-client.js ./src/generated/ClientSolApi.ts @moralisweb3/common-sol-utils"
```
ErnoW, line 19: Does not work anymore

packages/client/authUtils/src/AuthClientError.ts
```diff
@@ -0,0 +1,27 @@
+export enum AuthClientErrorCode {
+  GENERIC = 'CA0000',
+  ALREADY_AUTHENTICATED = 'CA0001',
+  NOT_AUTHENTICATED = 'CA0002',
+  ALREADY_CONNECTED = 'CA0003',
+  NOT_CONNECTED = 'CA0004',
+}
+
+// TODO: we don't use errors from the @moralisweb3/core package, because we would have a collision with the MoralisAuthError type.
```
ErnoW, line 9: Seems like a good reason to prioritise the structure of error codes. Ideally we want them to be in the specified packages. 👍

packages/client/authUtils/src/Connector.ts
```diff
@@ -0,0 +1,18 @@
...
+  provider: WalletProvider;
+  readWallet(): Promise<WalletDetails>;
+  signMessage(message: string): Promise<string>;
+  disconnect(): Promise<void>;
+}
+
+export interface WalletDetails {
+  address: string;
+  chain?: string;
+  network?: string;
```
ErnoW, line 17: Lets rename this networkType, as we had issues in the past where users confuse network with chain.

packages/client/backendAdapter/frontendOnly/src/FrontEndOnlyAuth.ts
```diff
@@ -0,0 +1,76 @@
...
+        networkType,
+        address: session.address,
+        // TODO: here we should call the Moralis API and receive the profileId.
+        profileId: `TODO_DO_NOT_USE_THIS_VALUE_${session.address}_TODO`,
+      };
+    }
+    return null;
+  }
+
+  public async getMessageToSign(request: GetMessageToSignRequest): Promise<GetMessageToSignResponse> {
```
ErnoW, line 31: This only signs the message without using our auth service, right?
We already have a task for implementing auth with frontendOnly adapter, when our auth serivice supports this

packages/client/backendAdapter/utils/src/NetworkTypeResolver.test.ts
```diff
@@ -0,0 +1,13 @@
+import { NetworkTypeResolver } from './NetworkTypeResolver';
```
ErnoW, line 1: Can you add test coverage reporting to our CI? 
in `main.yml` > `Test coverage comment`

The same holds for the other packages that have been added that include tests

ErnoW (changes requested):

demos/webpack/src/main.ts
b4rtaz: Nice idea! I'll add this.

packages/client/authUtils/src/Connector.ts
b4rtaz: Here `network` doesn't mean the network type. This is a Solana network. I'll rename it to `solNetwork` and add some comment.

packages/client/backendAdapter/frontendOnly/src/FrontEndOnlyAuth.ts
b4rtaz: > This only signs the message without using our auth service, right?

Yes.

When the API will be ready we can refactor this file to support the changes.

ErnoW (approved):

PR #909, opened 2022-12-20
@moralisweb3/express

Y0moo:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

### Solution Description

- Added new `@moralisweb3/express` package.
- Updated the `demos/express` demo
- Created codegen for `@moralisweb3/express`


.prettierignore
```diff
@@ -12,6 +12,7 @@ packages/**/generated/**/*
 **/generated/**/*
 packages/core/src/version.ts
 **/src/operations/openapi.ts
+!packages/express/src/**/generated/**/*
```
ErnoW, line 15: Why this exception? The generated code is not prettyfied, and we don't want that to influence the prettier check.

ErnoW (commented):
Looks good. Would be nice if we can have some testing in place for this if possible.

Also, these things will be picked up in the next issues right?
- Authentication handling with JWT
- Generated documentation of the available endpoints

.prettierignore
Y0moo: Why have we decided to exclude auto-generated code from the prettier formatting?

Y0moo:
> Looks good. Would be nice if we can have some testing in place for this if possible.
> 
> Also, these things will be picked up in the next issues right?
> 
> * Authentication handling with JWT
> * Generated documentation of the available endpoints

Yes, I think it'll be better to work on these two steps during client package development

.prettierignore
Y0moo: I will add some tests asap

ErnoW: > Why have we decided to exclude auto-generated code from the prettier formatting?

Because you should commit generated code directly. Otherwise when we re-generate we have a lot of changes that are only formatting-related. This is bad dev-experience and will result in issues if we run code generation in CI.

So we need to:
- Ingnore these files in Prettier
- Include prettier formatting as part of code generation

ErnoW (approved):

b4rtaz:
I wonder about one thing. Do we need to define those interfaces? Route handlers are only on the backend side and no one uses it.

```
type RequestParams = Pick<GetWalletTokenTransfersJSONRequest, 'address'>;

type RequestQuery = Pick<
  GetWalletTokenTransfersJSONRequest,
  'chain' | 'subdomain' | 'fromBlock' | 'toBlock' | 'fromDate' | 'toDate' | 'limit' | 'cursor'
>;
```

b4rtaz:
So maybe we could reduce an amount of files in this package, because *Resolver.files doesn't give any value than a generic resolver like:

```ts
export const genericResolver = async (
  operation: Operation,
  req: Request<RequestParams, undefined, undefined, RequestQuery>,
  res: Response,
  next: NextFunction,
) => {
   // we have a operation and we can parse a request in a general way.
};
```

Y0moo:
> I wonder about one thing. Do we need to define those interfaces? Route handlers are only on the backend side and no one uses it.
> 
> ```
> type RequestParams = Pick<GetWalletTokenTransfersJSONRequest, 'address'>;
> 
> type RequestQuery = Pick<
>   GetWalletTokenTransfersJSONRequest,
>   'chain' | 'subdomain' | 'fromBlock' | 'toBlock' | 'fromDate' | 'toDate' | 'limit' | 'cursor'
> >;
> ```

I've decided to define them so users can have a nice typescript support if they use `Resolvers`
```
apiRouter.get('/nft/:address/:tokenId/owners', evmGetNFTTokenIdOwnersResolver);
```

b4rtaz:
> I've decided to define them so users can have a nice typescript support if they use Resolvers

I see, but that still doesn't explain those types/interfaces. You could prepare only variables (like you want) and one generic resolver.

```ts
const xResolver = createResolver(xOperation);
const yResolver = createResolver(yOperation);
// ...
```

Y0moo:
> > I've decided to define them so users can have a nice typescript support if they use Resolvers
> 
> I see, but that still doesn't explain those types/interfaces. You could prepare only variables (like you want) and one generic resolver.
> 
> ```ts
> const xResolver = createResolver(xOperation);
> const yResolver = createResolver(yOperation);
> // ...
> ```

Oh, now I see what you mean. Yeah, nice idea, thank you! Let me check how it fits there

Y0moo:
Hey @b4rtaz! I've checked again your suggestion, and actually, it's not that simple, because a lot of "if" should be used if we select this design. At the moment resolvers are not just "simple generic resolver"s, you can see it from the handlebars templates. Also, we need to have at least three generic resolvers for basic, nullable and pagintated resolvers.

I agree there is a lot of space to make it more efficient, but redesigning an already working package will take time. So we can improve it in next sprints.

b4rtaz:
> So we can improve it in next sprints.

Ok, if this is not possible at this moment that's fine.

>  it's not that simple, because a lot of "if" should be used if we select this design

Yep. I think will be similar to [this file](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/main/packages/client/evmApi/src/generated/EvmApiClient.ts) and [that generator](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/main/packages/client/backendAdapter/utils/scripts/generate-client.js).

b4rtaz:
So basically the generator should generate something like:

```ts
const aResolver = createPaginatedResolver(aOperation);
const bResolver = createNullableResolver(bOperation);
const cResolver = createResolver(cOperation);
const dResolver = createNullableResolver(dOperation);
const eResolver = createResolver(eOperation);
...
```

Y0moo:
Could you approve the MR so we merge it? @b4rtaz

Y0moo:
> So basically the generator should generate something like:
> 
> ```ts
> const aResolver = createPaginatedResolver(aOperation);
> const bResolver = createNullableResolver(bOperation);
> const cResolver = createResolver(cOperation);
> const dResolver = createNullableResolver(dOperation);
> const eResolver = createResolver(eOperation);
> ...
> ```

Sounds like a nice plan

packages/express/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Resolver usage example:
+
+```ts
+import { evmGetNFTTokenIdOwnersResolver } from '@moralisweb3/express'
+
+export const apiRouter = express.Router();
+
+apiRouter.get('/nft/:address/:tokenId/owners', evmGetNFTTokenIdOwnersResolver);
+```
```
b4rtaz, line 28: Your package provides auth features like AuthRouter etc, but I don't see how I can protect my own controllers/HTTP handlers against not authenticated requests. Do we want to provide any gating solution? Or write some explanation?

```ts
expressRouter.get('/my-secret-endpoint', (req, res) => {
    // this is an endpoint for only authenticated users
})
```

b4rtaz (commented):
^

packages/express/README.md
Y0moo: JWT auth middleware is a part of backendAdapter PR. You can find it here: https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/beta-express-backend-adapter/packages/express/src/moralisJWT.ts

Y0moo: So its usage is like in the example bellow:
```
const moralisSessionChecker = moralisJWT(config.TOKEN_SECRET);

apiRouter.use('/evm-api', moralisSessionChecker, apiLimiter, EvmApiRouter.Router);
apiRouter.use('/solana-api', moralisSessionChecker, apiLimiter, SolApiRouter.Router);
```

packages/express/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Router usage example:
+
+```ts
+import { EvmApiRouter, SolApiRouter } from '@moralisweb3/express'
+
+export const apiRouter = express.Router();
+
+apiRouter.use('/evm', EvmApiRouter.Router);
+apiRouter.use('/solana', SolApiRouter.Router);
```
b4rtaz, line 17: If we want to connect the client-side package to the backend with this package, we should avoid defining route path at user's app level, because the user can set different one. So maybe we should have a more direct syntax, like:

```ts
apiRouter.use(moralisExpress.evmApi());
apiRouter.use(moralisExpress.solApi());
```

Of course it can be replaced.

```ts
apiRouter.use(moralisExpress.evmApi({
  url: '/my-custom-evm'
}));
```

b4rtaz (commented):
^

packages/express/README.md
b4rtaz: I see. But I mean, we should have a mention it in the documentation. I think users what will take our package, they don't want to hack the package. This feature is a basic one.

packages/express/README.md
Y0moo: Good point! I also thought about this so I decided to add a way for users to set module Url in backendAdapter configuration:
[```
https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/beta-express-backend-adapter/demos/express-fullstack-client/src/main.ts#L146
```](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/beta-express-backend-adapter/demos/express-fullstack-client/src/main.ts#L146)

Comment (Y0moo) [reply packages/express/README.md:17]: So users can define something like:
```
backendAdapter: ExpressBackendAdapter.create({
      baseUrl: 'http://127.0.0.1:4000/api',
      api: {
        solana: {
          relativePath: '/web3/moralis-api/solana'
        }
      }
    }),
```

Comment (Y0moo) [reply packages/express/README.md:17]: Your proposed solution is also nice, I guess it's better to use it here

Comment (Y0moo) [reply packages/express/README.md:17]: I love the zero config idea

Comment (Y0moo) [reply packages/express/README.md:17]: > If we want to connect the client-side package to the backend with this package, we should avoid defining route path at user's app level, because the user can set different one. So maybe we should have a more direct syntax, like:
> 
> ```ts
> apiRouter.use(moralisExpress.evmApi());
> apiRouter.use(moralisExpress.solApi());
> ```
> 
> Of course it can be replaced.
> 
> ```ts
> apiRouter.use(moralisExpress.evmApi({
>   url: '/my-custom-evm'
> }));
> ```

Looks like we can not make it without overengineering, because `.use` accept URL as a first param. We still can use `apiRouter.use(moralisExpress.evmApi());`, but it becomes harder to add middlewares:
```
apiRouter.use(rateLimitter, moralisExpress.evmApi());
```
because the URL path is not defined.

Comment (Y0moo) [reply packages/express/README.md:17]: We can add config with `middleWares`, but for me it looks like making thing harder, not simplier

Comment (Y0moo) [reply packages/express/README.md:17]: What do you think?: 
```
apiRouter.use(EvmApi.path, EvmApi.router);
```

Comment (b4rtaz) [reply packages/express/README.md:17]: > Looks like we can not make it without overengineering, because .use accept URL as a first param. 

There is another method overloading. Check [serve-static](https://github.com/expressjs/serve-static/blob/master/index.js) package. I can use it in this way:

```ts
apiRouter.use(express.static(...))
```

(`express.static` is an [alias](https://github.com/expressjs/express/blob/508936853a6e311099c9985d4c11a4b1b8f6af07/lib/express.js): `exports.static = require('serve-static');`)

Y0moo:
Hey @b4rtaz! I've refactored the code with the approach you proposed with generic resolvers. Please review and let me know what do you think?

PR #921, opened 2022-12-23
Auth MagicLink connector

FedericoAmura:
---
name: 'Auth MagicLink connector'
about: This PR adds a connector to use MagicLink in Moralis Client Auth 
---

### Checklist

### Issue Description


packages/client/connector/magicLink/src/MagicLinkEvmConnector.ts
```diff
@@ -0,0 +1,64 @@
...
+    };
+  }
+
+  public async signMessage(message: string): Promise<string> {
+    const signer = this.provider.getSigner();
+    return signer.signMessage(message);
+  }
+
+  public async disconnect(): Promise<void> {
+    // @ts-ignore
```
b4rtaz, line 60: `@ts-ignore` is a very bad practice. It can hide a serious error.

packages/client/connector/magicLink/src/MagicLinkEvmConnector.ts
```diff
@@ -0,0 +1,64 @@
...
+      extensions: [new ConnectExtension(), ...(options.extensions ?? [])],
+    };
+
+    this.magic = new Magic(apiKey, this.options);
+  }
+
+  public async connect(): Promise<EvmConnection> {
+    const provider = this.magic.rpcProvider;
+    await provider.enable();
+    // @ts-ignore
```
b4rtaz, line 31: `@ts-ignore` - bad practice.

demos/webpack/README.md
```diff
@@ -2,6 +2,12 @@
```
b4rtaz, line 2: I don't see updated `demos/webpack/package.json` file. You should add a dependency to the magic link package.

b4rtaz (changes requested):
Nice work, but I've found a few scratches. The build crashes as well.

I wonder about restoreability of a provider, can I authenticate -> reload page -> restore provider?

demos/webpack/README.md
```diff
@@ -2,6 +2,12 @@
 
 This project contains a simple webpack app.
 
+## 🪄 Auth with MagicLink
+
+1. Create a [Magic](https://magic.link/) account.
+2. Create a Magic Connect App.
+3. Copy the PUBLISHABLE API KEY and paste it in place of `MAGIC_CONNECT_PUBLISHABLE_API_KEY`.
+
```
b4rtaz, line 10: We could move all secret keys to a `.env` file (+ `.env.example`).

b4rtaz (changes requested):

packages/client/connector/magicLink/src/MagicLinkEvmConnector.ts
FedericoAmura: They haven't fixed that compatibility, but changing that to an `any` casting as in the Metamask one [(it is also the kind of solution they give)](https://github.com/magiclabs/magic-js/issues/126)

demos/webpack/README.md
FedericoAmura: Added

demos/webpack/README.md
FedericoAmura: I could not achieve this. However I extracted it to a constant at the top, and made the buttons hide when this is not configured

This is a FE app, so basically there is nothing secret. I know React has env file support but I haven't managed to configure webpack to give us the same feature in this demo. I struggled with several errors trying to get it and don't think this is worth that much time, at least we can consider that as a follow up

b4rtaz: > This is a FE app, so basically there is nothing secret.

This is not a point. This is a demo for customers, so it should be easy to use. Check out our other demos.

packages/client/connector/magicLink/src/MagicLinkEvmConnector.ts
b4rtaz: I think casting to `any` is a better solution than `@ts-ignore`. `@ts-ignore` hides obvious errors.

b4rtaz (changes requested):

packages/client/connector/magicLink/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "devDependencies": {
+    "prettier": "^2.5.1",
+    "typescript": "^4.9.3"
+  },
+  "dependencies": {
+    "@babel/core": "^7.20.5",
+    "@babel/plugin-transform-react-jsx": "^7.19.0",
+    "@magic-ext/connect": "^4.0.0",
+    "@moralisweb3/client-auth-utils": "workspace:^",
+    "@moralisweb3/client-evm-auth": "workspace:^",
```
b4rtaz, line 27: There cannot be `workspace:^`. Because if we will publish that package, on a user pc that couldn't be resolved. Here you must put the version from the repo.

demos/webpack/src/main.ts
```diff
@@ -1,8 +1,12 @@
 import { MoralisClient } from '@moralisweb3/client';
 import { FrontEndOnlyBackendAdapter } from '@moralisweb3/client-backend-adapter-frontend-only';
+import { EvmConnector } from '@moralisweb3/client-evm-auth';
 import { Connection, clusterApiUrl, PublicKey } from '@solana/web3.js';
+import { MagicLinkEvmConnector } from '@moralisweb3/client-connector-magic-link';
 import { WalletConnectEvmConnector } from '@moralisweb3/client-connector-wallet-connect';
 
+const MAGIC_CONNECT_PUBLISHABLE_API_KEY = '';
```
b4rtaz, line 8: It's still a constant. Some tip: you need to install the `dotenv` package to the webpack demo to support the `.env` file.

b4rtaz (changes requested):

demos/webpack/src/main.ts
FedericoAmura: Did that, only got more errors and finally got stuck. Tried also with `webpack-dotenv` package and still no success. Gonna try again today with some rest

packages/client/connector/magicLink/src/MagicLinkEvmConnector.ts
FedericoAmura: I'm on it. Magic typing system is complex as it is based on extensions. Worst case, I'll have to type it myself

packages/client/connector/magicLink/package.json
FedericoAmura: Missed it. Great catch
Changed

demos/webpack/src/main.ts
FedericoAmura: Fixed this. I could bet I did the same thing yesterday but it didn't worked...

b4rtaz (changes requested):
I'm not sure that we should add the react demo now. We want to create a react package with dedicated syntax for react (similar to nextjs).

demos/webpack/src/main.ts
```diff
@@ -106,18 +118,44 @@ async function onButtonClicked(e: Event) {
...
+      rpc: {
+        1: 'https://replace_me/',
+      },
+    }),
+  ];
+  if (MAGIC_CONNECT_PUBLISHABLE_API_KEY) {
+    // Add magic link auth button
+    const authenticateButtonBar = document.getElementById('authenticate')!;
+    const magicLinkAuthButton = document.createElement('button');
+    magicLinkAuthButton.setAttribute('class', 'button');
```
b4rtaz, line 132: This `if` is in a wrong place. Because if I won't set the `MAGIC_CONNECT_PUBLISHABLE_API_KEY` variable, and can click at the button and I don't get a feedback that, there is something missing.

b4rtaz (changes requested):

FedericoAmura:
> I'm not sure that we should add the react demo now. We want to create a react package with dedicated syntax for react (similar to nextjs).

Good point. I'll remove it once I'm finished. I added this to be able to debug all this client code anyway

demos/webpack/src/main.ts
FedericoAmura: If you don't set `MAGIC_CONNECT_PUBLISHABLE_API_KEY` that button will not be shown because it is added on line 137 `authenticateButtonBar.appendChild(magicLinkAuthButton);` (same with the connect button)

But you are right in the goal, and I'm fixing that too for wallet connect buttons. Probably it will be better if I only disable the buttons and show an alert so users know they can do that. What do you think?

b4rtaz: I think the best would be to show all buttons and displays an error if the connector is not configured. In this way a user sees all possibilities at start.

FedericoAmura: Done that way

b4rtaz (approved):

PR #926, opened 2022-12-29
renamed block_hash to hash

b4rtaz:
…hod.

---
name: renamed block_hash to hash
---

### DIFF
--- a/.changeset/curvy-bugs-protect.md
+++ b/.changeset/curvy-bugs-protect.md
@@ -1,0 +1,5 @@
+---
+'@moralisweb3/common-evm-utils': patch
+---
+
+Renamed the `block_hash` field to `hash` for the response of the `getDateToBlock()` method (according to the API changes).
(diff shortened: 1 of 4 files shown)


.changeset/curvy-bugs-protect.md
```diff
@@ -0,0 +1,5 @@
+---
+'@moralisweb3/common-evm-utils': patch
+---
+
+Renamed the `block_hash` field to `hash` for the response of the `getDateToBlock()` method (according to the API changes).
```
FedericoAmura, line 5: Considering this removes the `block_hash` property, I would consider this a breaking change. We should keep the previous one as deprecated or consider a major version bump

b4rtaz: Good point!

ErnoW: It's a breaking change on the api part. At the moment renaming params on api are common, and releasing a major bump on these cases is too much in my opinion.

I agree with a deprecated fallback on the result 👍

b4rtaz: To add the `depreciated` attribute I needed to add a dedicated type (`EvmBlockDate`). Additionally I've renamed the rest fields to camelCase.

ErnoW (approved):

PR #953, opened 2023-01-09
fix: serialisation nextjs requests

ErnoW:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

JSONs get double serialised before sending it to the api without deserialising the request

### Solution Description

- Serialise requests only once
- Deserialise request before sending it to the api


packages/next/src/MoralisNextApi/Modules.ts
```diff
@@ -0,0 +1,60 @@
...
+export class Module {
+  public name: string;
+  public baseUrl: string;
+  public operations: UnknownOperation[];
+
+  constructor({ name, baseUrl, operations }: { name: string; baseUrl: string; operations: UnknownOperation[] }) {
+    this.name = name;
+    this.baseUrl = baseUrl;
+    this.operations = operations;
+  }
```
b4rtaz, line 21: This could by simplify by using the property promotion feature:

```ts
public constructor(
    public readonly name: string,
    public readonly baseUrl: string,
    public readonly operations: UnknownOperation[],

);
```

packages/next/src/MoralisNextApi/Modules.ts
```diff
@@ -0,0 +1,60 @@
+import { EvmApi } from '@moralisweb3/evm-api';
+import { operations as evmOperations } from 'moralis/common-evm-utils';
+import { operations as solOperations } from 'moralis/common-sol-utils';
+import { Operation } from 'moralis/common-core';
+import { operations as authOperations } from '@moralisweb3/common-auth-utils';
+import { Auth } from '@moralisweb3/auth';
+import { SolApi } from '@moralisweb3/sol-api';
+import Moralis from 'moralis';
+
+export type UnknownOperation = Operation<unknown, unknown, unknown, unknown>;
```
b4rtaz, line 10: Maybe we should move this type to the core package (`packages/common/core/src/operations/Operation.ts`). We have many places in the repo where we define the `UnknownOperation` type.

packages/next/src/MoralisNextApi/MoralisNextApi.ts
```diff
@@ -3,22 +3,23 @@ import { RequestHandlerResolver } from './RequestHandlerResolver';
...
+async function MoralisNextHandler({ req, res, authentication, core }: MoralisNextHandlerParams) {
   const [moduleName, operationName] = req.query.moralis as string[];
 
   try {
-    const requestHandler = RequestHandlerResolver.tryResolve(moduleName, operationName);
-    if (!requestHandler) {
-      return res.status(500).json({ error: `Operation ${moduleName}/${operationName} is not supported` });
-    }
+    const module = getModuleByName(moduleName);
+    const operation = module?.getOperationByName(operationName);
```
b4rtaz, line 13: Not needed question mark (`?`).

b4rtaz (changes requested):

packages/next/src/MoralisNextApi/Modules.ts
ErnoW: The interface is better readable when providing these values as object properties (at least when having more than 2). I don't think this syntax works when doing it like that, as far as I know. I want to keep the interface as:
```
{
    name: EvmApi.moduleName,
    baseUrl: Moralis.EvmApi.baseUrl,
    operations: evmOperations as UnknownOperation[],
}
```

packages/next/src/MoralisNextApi/Modules.ts
ErnoW: Good one 👍

b4rtaz (dismissed):

packages/next/src/MoralisNextApi/Modules.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Auth } from '@moralisweb3/auth';
+import { SolApi } from '@moralisweb3/sol-api';
+import Moralis from 'moralis';
+
+export class Module {
+  public name: string;
+  public baseUrl: string;
+  public operations: UnknownOperation[];
+
+  constructor({ name, baseUrl, operations }: { name: string; baseUrl: string; operations: UnknownOperation[] }) {
```
FedericoAmura, line 15: The parameter should be interface Module?
We can use Required<Module> or something similar to not use the class and derive a type

b4rtaz: Do you suggest something like?

```ts
interface ModuleDescriptor {
    name: string;
    baseUrl: string;
    operations: UnknownOperation[];
}

class Module {

public constructor(private readonly descriptor: ModuleDescriptor) {

}
```

I think this is better than the current approach. The current approach wastes the memory only for a better readability.

FedericoAmura: No, I just wanted to type the constructor input

```
type ModuleType = Required<Module>;

class Module {
    name: string;
    baseUrl: string;
    operations: UnknownOperation[];

    public constructor({ name, baseUrl, operations }: ModuleType) {
        this.name = name;
        this.baseUrl = baseUrl;
        this.operations = operations;
    }
}
```


That approach wouldn't be vulnerable to this?:
```
const descriptor = { name: 'aModule', baseUrl: 'https://myurl.com', operations: [] };
const module = new Module(descriptor);

descriptor.baseUrl = 'https://someOtherUrlThatIsNotReadonly';
// And now the module has a different baseUrl because the `readonly` only affects `descriptor`
```
I would prefer an object to not depend anymore on the things it received in its constructor if possible

ErnoW:
@b4rtaz @FedericoAmura I continued on fixing another bug, included it in this PR as it relates to the previous discussion.

This change allow us to remove the number of arguments to 2, so I removed the object interface now. I do agree with the suggestion @FedericoAmura to have a typed interface before implementing (but it is not needed anymore now).

See commit https://github.com/MoralisWeb3/Moralis-JS-SDK/pull/953/commits/cf2a8e36e277b543cd61eb9a0d10cb496afbc559 for changes related to the other bug. It is about that we do not account for Nullable/Paginated resolvers. Now we resolve the operations based on the EvmApi/SolApi/Auth definitions so that we have a single source of truth for the resolving logic.

b4rtaz (approved):

packages/next/src/MoralisNextApi/MoralisNextApi.ts
```diff
@@ -50,7 +54,8 @@ const MoralisNextApi = ({ authentication, ...config }: MoralisNextApiParams) =>
     Moralis.start(config);
   }
 
-  return async (req: NextApiRequest, res: NextApiResponse) => MoralisNextHandler({ req, res, authentication });
+  return async (req: NextApiRequest, res: NextApiResponse) =>
+    MoralisNextHandler({ req, res, authentication, core: Moralis.Core });
```
Y0moo, line 58: Why do we pass here the "core" argument? The Moralis instance is the same across the whole application, so we can import it into any part of the application instead of passing it as an argument. @ErnoW

ErnoW: It's a better design pattern to include these dependencies as argument (see [Dependency injection](https://en.wikipedia.org/wiki/Dependency_injection))

In this case it does not make too much of a difference as in NextJs we also use directly imported instances of the 'core'. But it's a good habit to include it as a dependency wherever we can. But it results in more predictable code.

This is also why we pass around the `core` instance within all the modules/functions within the SDK.

PR #955, opened 2023-01-10
Feat walletconnect auth

FedericoAmura:
---
name: 'Wallet Connect V2 auth'
about: This PR adds support for a Wallet Connect V2 connector in Moralis Auth and also adds listeners for when the user revokes the connection from the wallet instead of the app
---

### Checklist

### Issue Description


b4rtaz (changes requested):
I'm not sure that we want to keep both versions of Wallet Connect. I think we should simplify the user experience, and this introduces ambiguity.

FedericoAmura:
> I'm not sure that we want to keep both versions of Wallet Connect. I think we should simplify the user experience, and this introduces ambiguity.
> 
> Or maybe, there is some important reason to keep both.

For what I tested until now, there are several wallets that don't support WC2. And it is a little bit harder to configure and use properly. I would suggest to keep both and document that you should go with V2, and if for some reason you have to use V1 (maybe you want to use an unsupported wallet), it is there too.

Anyway I'll be testing more wallets and revisiting the config today, so hopefully I will be able to get around those things

PR #966, opened 2023-01-16
fix: resolving requestChallenge hooks

ErnoW:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

After changing the logic of resolving operations in  NextJs, we have an issue. Auth is handled a bit different, as we don't export the resolvers directly, but rather a combined function.

Two possible solutions:
1. Change the NextJs hook to be compatible with `requestMessage` and pass the network as a parameter. Then we need to update `packages/next/src/MoralisNextApi/Modules.ts` to account for a "different kind" of operation resolver (not grouped)
2. Add additional groups/operations to Auth, that is compatible with swagger/our operations structure. 

I choose option 2 as it is more predictable and does not involve a breaking change in the NextJs package. In the future we could improve it by extending Auth from a ClientEvmAuth and ClientSolAuth (similar to what we do in EvmApi). This requires restructuring of the Auth make it able to extend multiple other "client" classes. I have omitted it for now as we are also in the discussions of making it easier to generate our Api modules, based on swagger (for the current beta).

### DIFF
--- a/.changeset/forty-dragons-act.md
+++ b/.changeset/forty-dragons-act.md
@@ -1,0 +1,6 @@
+---
+'@moralisweb3/auth': patch
+'@moralisweb3/next': patch
+---
+
+Fix auth hooks for NextJs
--- a/.changeset/forty-dragons-act.md
+++ b/.changeset/forty-dragons-act.md
@@ -1,6 +1,5 @@
 ---
 '@moralisweb3/auth': patch
-'@moralisweb3/next': patch
 ---
 
 Fix auth hooks for NextJs
--- a/packages/auth/src/Auth.ts
+++ b/packages/auth/src/Auth.ts
@@ -4,7 +4,14 @@ import { makeVerify, VerifyEvmOptions, VerifyOptions, VerifySolOptions } from '.
 import {
   VerifyChallengeSolanaResponseAdapter,
   VerifyChallengeEvmResponseAdapter,
+  RequestChallengeEvmResponseAdapter,
+  requestChallengeEvmOperation,
+  RequestChallengeEvmRequest,
+  RequestChallengeSolanaResponseAdapter,
+  requestChallengeSolanaOperation,
+  RequestChallengeSolanaRequest,
 } from '@moralisweb3/common-auth-utils';
+import { OperationResolver } from '@moralisweb3/api-utils';
 
 export const BASE_URL = 'https://authapi.moralis.io';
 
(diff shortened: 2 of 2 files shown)


packages/auth/src/Auth.ts
```diff
@@ -27,6 +34,23 @@ export class Auth extends ApiModule {
     // Nothing
   }
 
+  // Client-side compatible operation, structured in a predictable way as defined in the operation
+  // TODO: generate in seperate package "client-evm-auth" (similar to client-evm-auth)
```
b4rtaz, line 38: `// TODO: generate in seperate package "client-evm-auth" (similar to client-evm-auth)`

Should be there `generate in seperate package "_____server_____-evm-auth"`?

.changeset/forty-dragons-act.md
```diff
@@ -0,0 +1,6 @@
+---
+'@moralisweb3/auth': patch
+'@moralisweb3/next': patch
+---
+
+Fix auth hooks for NextJs
```
b4rtaz, line 6: I see in this PR only changes in the `auth` package. Where are changes for the `@moralisweb3/next` package?

b4rtaz (changes requested):

.changeset/forty-dragons-act.md
ErnoW: You're right, after cleaning up no changes in NextJs were needed anymore 👍

packages/auth/src/Auth.ts
ErnoW: I think it should be the same as EvmApi, where it extends upon ClientEvmApi.

In this case, we only want to support requestMessage operations to be available in the client though.

Or:
- we need to opt for another strategy where we define client-side support per operation. 
- allow all operations in the client-side (at least the definitions), and let the API block any requests that should not happen from client-side (configured in the access control for the new API keys)

b4rtaz: But the `ClientEvmApi` doesn't use/extend the `EvmApi` class. It uses directly [the operations](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/beta/packages/client/evmApi/src/generated/EvmApiClient.ts).

For the client packages (currently) we don't give a possibility to call `requestChallenge` or `verify` methods manually. These method are wrapped by the `backendAdapter` layer.

b4rtaz: \+ The client auth packages (`@moralisweb3/client-evm-auth`, `@moralisweb3/client-sol-auth`) wont't use the `Auth` class internally. To be more precise we talk about the `@moralisweb3/client-backend-adapter-frontend-only` package. That package will use directly the operations as well.

b4rtaz (approved):

PR #989, opened 2023-01-26
feat: Moralis Streams in Aptos

FedericoAmura:
---
name: 'Support for Aptos Streams'
about: This PR adds support for Moralis Streams in Aptos
---

### Checklist

### Issue Description


packages/common/aptosUtils/src/dataTypes/AptosAddress/AptosAddress.test.ts
```diff
@@ -0,0 +1,60 @@
...
+
+      expect(aptos.address).toEqual(address);
+      expect(aptos.toJSON()).toEqual(address);
+      expect(aptos.toString()).toEqual(address);
+      expect(aptos.format()).toEqual(address);
+    });
+  }
+
+  it('create() throws an error when a passed address is invalid', () => {
+    expect(() => AptosAddress.create('invalid')).toThrowError('[C0005] Invalid address provided: invalid');
```
ErnoW, line 26: Can we have a few tests for a few addresses similar to a valid address that are invalid? For example invalid length etc.

packages/common/aptosUtils/src/dataTypes/AptosAddress/AptosAddress.test.ts
```diff
@@ -0,0 +1,60 @@
+import { AptosAddress } from './AptosAddress';
+
+describe('AptosAddress', () => {
+  const ADDRESS = '0xeeff357ea5c1a4e7bc11b2b17ff2dc2dcca69750bfef1e1ebcaccf8c8018175b'; // account
+
+  const addresses = [
```
ErnoW, line 6:
```suggestion
  const validAddresses = [
```

packages/common/aptosUtils/src/dataTypes/AptosAddress/AptosAddress.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    ADDRESS,
+    '0x19aadeca9388e009d136245b9a67423f3eee242b03142849eb4f81a4a409e59c',
+    '0000357ea5c1a4e7bc11b2b17ff2dc2dcca69750bfef1e1ebcaccf8c8018175b', // no 0x
+    '0x357ea5c1a4e7bc11b2b17ff2dc2dcca69750bfef1e1ebcaccf8c8018175b', // no leading 0s
+    '0x0000357ea5c1a4e7bc11b2b17ff2dc2dcca69750bfef1e1ebcaccf8c8018175b',
+  ];
```
ErnoW, line 12: I assume that Aptos addresses are always lowercase? It is not like with EVM that there are different formats (checksum [EIP-55](https://eips.ethereum.org/EIPS/eip-55))

packages/common/streamsUtils/package.json
```diff
@@ -33,6 +33,7 @@
   },
   "dependencies": {
     "@ethersproject/abi": "^5.7.0",
+    "@moralisweb3/common-aptos-utils": "workspace:^",
```
ErnoW, line 36: Never seen this before, what does it mean, will it pull the version from the workspace? And will it still work when we publish the package, when it has no context of the workspace?

Should we not just use the 2.12.0 versioning, and let changeset handle the versioning?

packages/common/streamsUtils/src/operations/aptosStreams/updateStreamStatusAptosOperation.ts
```diff
@@ -0,0 +1,81 @@
+import { Camelize, Operation, ResponseAdapter } from '@moralisweb3/common-core';
+import { operations } from '../openapi';
+
+type OperationId = 'aptosStreamsUpdateStatus';
+
+type PathParams = operations[OperationId]['parameters']['path'];
+// TODO openapi spec has an impossible type for this operation (it should be UNION, not INTERSECTION)
```
ErnoW, line 7: Did you report this issue already?

packages/common/streamsUtils/src/operations/evmStreams/deleteStreamEvmOperation.ts
```diff
@@ -55,8 +54,8 @@ function getRequestBody() {
   return {};
 }
 
-function deserializeResponse(jsonResponse: DeleteStreamEvmJSONResponse, request: DeleteStreamEvmRequest, core: Core) {
-  return EvmStream.create(jsonResponse, core);
```
ErnoW, line 59: Why was this change made? It is a breaking change

packages/streams/src/methods/addAddress.ts
```diff
@@ -3,26 +3,36 @@ import { OperationResolver } from '@moralisweb3/api-utils';
...
   const evmFetcher = new OperationResolver(addAddressEvmOperation, baseUrl, core).fetch;
 
-  return ({ networkType, ...options }: AddAddressOptions): Promise<AddAddressEvmResponseAdapter> => {
+  return ({
+    networkType,
+    ...options
+  }: AddAddressOptions): Promise<AddAddressAptosResponseAdapter | AddAddressEvmResponseAdapter> => {
     switch (networkType) {
+      case StreamNetwork.APTOS:
+        return aptosFetcher({ ...(options as AddAddressAptosRequest) });
```
ErnoW, line 30: I really dislike the use of `as` here as it can result to Typescript errors. Can we not infer the type based on the provided options (like we do in the `Auth` methods)?

ErnoW (changes requested):

packages/common/aptosUtils/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@moralisweb3/common-aptos-utils",
```
b4rtaz, line 2: I'm not sure, but we have a naming convention that the name of network should have 3 characters (`@moralisweb3/evm-utils`, `@moralisweb3/sol-utils`). Should be there `common-apt-utils`?

b4rtaz (changes requested):

packages/common/aptosUtils/package.json
ErnoW: The convention of using 3 characters is bad design nherited from v1. I prefer full names. Good point @b4rtaz as we should reconsider the naming as wel in v3. I prefer to have it evm, solana and aptos eventually. We should only deviate from this for long names or when the acronym is used in every day language (like EVM).

For this reason I prefer to use aptos as full name in this PR as well

packages/common/streamsUtils/src/operations/evmStreams/deleteStreamEvmOperation.ts
FedericoAmura: Reverted

packages/common/streamsUtils/package.json
FedericoAmura: Fixed. It was because of my environment

packages/common/streamsUtils/src/operations/aptosStreams/getAddressesAptosOperation.ts
```diff
@@ -0,0 +1,88 @@
...
+// Exports
+
+export interface GetAddressesAptosRequest extends Camelize<Omit<RequestParams, 'limit'>> {
+  limit?: number;
+}
+
+export type GetAddressesAptosJSONRequest = ReturnType<typeof serializeRequest>;
+
+// export type GetAddressesAptosJSONResponse = SuccessResponse;
+// TODO temporary fix. The openapi spec says that the response is an array of strings, but the actual response is a paginated array of objects with a single string property called "address"
```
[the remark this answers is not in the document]
FedericoAmura: Fixed

packages/common/aptosUtils/src/dataTypes/AptosAddress/AptosAddress.test.ts
FedericoAmura: For what I found and tested, it doesn't care about casing.

packages/common/aptosUtils/src/dataTypes/AptosAddress/AptosAddress.test.ts
FedericoAmura: Added test for too long. Too short completes with leading 0s. There is already an invalid address test. I think that covers all possible cases

packages/common/aptosUtils/src/dataTypes/AptosAddress/AptosAddress.test.ts
FedericoAmura: Renamed

packages/common/streamsUtils/src/operations/aptosStreams/updateStreamStatusAptosOperation.ts
FedericoAmura: Fixed

packages/streams/src/methods/addAddress.ts
FedericoAmura: Agree. Consider that an iteration as I finally arrived at a solution that can work. It still uses `as` but differently and now it also defines the type of response you will get. This way, when you pass `networkType = 'evm'` typescript knows it will return and EvmStream and not make you cast it, removing this cast can work but don't have this feature. The same goes for aptos, and can be extended to solana or other blockchains we will add support

I tried several other approaches but this was the best one by far. Destructuring params make the type system complex. Also, I like more the `<Module>.evm.<endpoint>` approach but that would be a breaking change, I would consider that for next version

b4rtaz (approved):

ErnoW (approved):

PR #990, opened 2023-01-27
Feat: @moralisweb3/react

Y0moo:
---
name: 'Add new library @moralisweb3/react'
about: This PR adds new @moralisweb3/react library
---

### Checklist

### Issue Description


packages/react/src/hooks/resolvers/_useResolverNullable.ts
```diff
@@ -0,0 +1,40 @@
...
+import { useCallback } from 'react';
+import useSWR from 'swr';
+
+export function _useResolverNullable<Request, JSONRequest, Response, JSONResponse>(
+  operation: Operation<Request, JSONRequest, Response, JSONResponse>,
+  baseUrl: string,
+  request?: Request,
+) {
+  const { core, swrConfig } = _useClient();
+  const { fetch: resolve } = new NullableOperationResolver(operation, baseUrl, core);
```
ErnoW, line 13: Will this not result in issues on re-renders? As we are re-creating new operationResolvers every re-render. Probably a useCallback would be better

packages/react/src/hooks/resolvers/_useResolverNullable.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const fetcher = async (_url: string, req: Request) => {
```
ErnoW, line 15: Same as previous comment, it might be okay in this case. But I'm always hesitant to define functions inside a hook that are not wrapped by a useCallback, as it can easily result in re-render loops or performance issues

ErnoW (changes requested):
Overall looks good to me. The main concern I have is regarding the swr config.
Ideally we want to let the user set config:
- per request (this is important)
- potentially a default via the context

Also ideally we want to have an abstraction of using swr. As in the future we might need/want to swap out swr with another library, or they might introduce breaking changes that result in breaking changes for us.  Abstracting this dependency makes us a lot more flexible to extend/change behaviour.

Ideally we want to:
- Specify the interface that we want to expose to our user
- Implement third-party library as swr to help us implement this
- Define a 'wrapper' that is utilised by all fetch requests, and passes it to swr (and other configuration that we set ourself)
- Use this 'wrapper' for all of our requests.

An example of this is when you look at wagmi. They use react-query, and are somewhat coupled to it. But they preserve a lot of flexibility by utilising one/two base query hooks ([useQuery](https://github.com/wagmi-dev/wagmi/blob/main/packages/react/src/hooks/utils/query/useQuery.ts) and [useBaseQuery](https://github.com/wagmi-dev/wagmi/blob/main/packages/react/src/hooks/utils/query/useBaseQuery.ts)).

Maybe it is unavoidable for us to be tightly coupled, but lets at least make this abstraction in 1 place (like a _useSwr hook). What do you think?

packages/react/src/context/MoralisProvider.tsx
```diff
@@ -0,0 +1,51 @@
+import { Core } from 'moralis/common-core';
+import Moralis from 'moralis';
+import { MoralisConfigValues } from 'moralis/lib/config/MoralisConfig';
+import React, { useContext, useMemo } from 'react';
+import { SWRConfiguration } from 'swr';
+
+export interface MoralisContextValue {
+  core: Core;
+  swrConfig?: FetchConfig;
```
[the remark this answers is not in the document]
Y0moo: Thanks! When I designed the library I decided having only config at Provider config is enough. Now, I'm fully agreed with you. Also good point about fallbacks, this functionality is very important and I'll add it for sure. And actually, we have similar functionality on `next` package, don't know why I didn't add it here.

What I'll do:
1. Concat the `fetchConfig` into general `config` in `MoralisProvider`, so there will be only one config. And I believe it absolutely makes sense because it'll be easier for some users to set some general rules on the highest moralis library lvl.
2. Add fallbacks like `onSuccess`/`onError`... and fetch options as params to all our hooks. So it'll look like:
```
const { fetch: getBlock } = useEvmBlock({ 
  blockNumberOrHash: '10000',
  onSuccess: () => console.log("Success"),
  revalidateOnFocus: true,
});
```

packages/react/src/hooks/resolvers/_useResolverNullable.ts
Y0moo: Well, I don't think it makes sense here to add `useCallback`, it'll actually lead to worse optimization since it'll create a new instance of `NullableOperationResolver` every rerender(useCallback doesn't save result of calculations, it saves only a reference and will call all code execuition inside each rerender). But adding `useMemo` with `[operation, baseUrl, core]` can be a good solution here.

packages/react/src/hooks/resolvers/_useResolverNullable.ts
Y0moo: Btw, just noticed I forgot to add `request` dependency to `fetch` function

packages/react/src/hooks/resolvers/_useResolverNullable.ts
ErnoW: useCallback and useMemo are very similar. Both could be used, depends on the rest of the code. Both of them should not be called more than once if you specify `operation, baseUrl, core` (because those values should also not change, in a sensible structure)

Another solution could be to use useRef, as operation, baseUrl and core should not change at all, they are all pre-determined and should be only set once.

packages/react/src/hooks/resolvers/_useResolverNullable.ts
ErnoW: I'm not sure what you mean with worse optimization and "No variable dependencies"? If you wrap a function inside a useCallback, then it **only** gets updated when one of the dependencies change. Right now it **always** gets updated whenever a re-render triggers.

The only reason not to wrap it in a useCallback/useMemo is when you want to optimise for performance and you know that it is more performant to redefine the variable/function instead of calculating the equality of dependencies.

But in my experience I prefer to wrap most values inside a useMemo and useCallback, and opt-out of this behaviour when needed. As it is really really easy to otherwise have performance issues that are difficult to track down.

Y0moo: Added `useCallback` here

demos/react/src/index.tsx
```diff
@@ -0,0 +1,26 @@
+import ReactDOM from 'react-dom/client';
+import App from './App';
+import reportWebVitals from './reportWebVitals';
+import { MoralisProvider } from '@moralisweb3/react';
+import { BrowserRouter } from 'react-router-dom';
+import { ChakraProvider } from '@chakra-ui/react';
+
+const root = ReactDOM.createRoot(document.getElementById('root') as HTMLElement);
+
```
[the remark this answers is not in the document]
Y0moo: Now it's possible to set the `revalidateOnMount` option to not fetch data on component mount [example](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/react/demos/react/src/views/Home.tsx#L43)

Buy [default](https://github.com/MoralisWeb3/Moralis-JS-SDK/blob/react/packages/react/src/context/MoralisProvider.tsx#L36) `revalidateOnFocus` is `false` now.

packages/react/src/context/MoralisProvider.tsx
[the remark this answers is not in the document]
Y0moo: > * Avoid the use of `swr` (some people don't know SWR, and it's a better design to avoid names that are third-party specific. This makes us more flexible to change the interface/library in the future if needed. Ideally we would have a wrapper/utility that handles all interaction with SWR)

`swrConfig` name here is only internal, so normally users will not see it

packages/react/src/context/MoralisProvider.tsx
```diff
@@ -0,0 +1,69 @@
+import Moralis from 'moralis';
+import React, { FC, createContext, useContext, useMemo } from 'react';
+import { FetchConfig, MoralisConfig, MoralisContextValue, MoralisProviderProps } from './types';
+
+export const MoralisContext = createContext<MoralisContextValue | null>(null);
+
+export const _useClient = () => {
```
FedericoAmura, line 7: I'm not a fan of this style. It gives a false sense of privacy that JS doesn't really have. And in this case it can also break the checks that validate using hooks correctly (in renders or other hooks, no ifs, no loops)

We can be tempted of changing them "because they are private" without thinking it too much, and breaking our users code
I think it is far better to know they are public and treat a breaking change in them as that, a breaking change

Y0moo: Interesting point, thanks @FedericoAmura! 
> And in this case it can also break the checks that validate using hooks correctly (in renders or other hooks, no ifs, no loops)

Could you please explain it to me better?

What'd you suggest to use here?

b4rtaz: I think the main goal here is to prevent hardcoding one instance of a service in a whole application. What if a user want to use two instances (for example: with two different API keys)?

By this architecture (plus some adjustments) it's possible:

```tsx
<MoralisProvider name="a" apiKey="A">
<MoralisProvider name="b" apiKey="B">

   const a = useClient("a");
   const b = useClient("b");

  const [ data1 ] = useEvmBlock({ ... }, a);
  const [ data2 ] = useEvmBlock({ ... }, b);

</B.MoralisProvider>
</A.MoralisProvider>
```

The example with two different API keys is closest to our problem, but this is a common problem.

Y0moo: Well, it doesn't make sense to me. If users want to use different API keys in different parts of code why would they wrap whole application? The still allowed to wrap separate parts of code(e.g. pages) with own context

Y0moo: Code example provided by @b4rtaz. Looks to me like bad app design.

b4rtaz: I don't agree. By the context you can apply the [dependency inversion principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle) to the app (check: [Dependency Inversion in React](https://medium.com/@vittorioguerriero/dependency-inversion-in-react-58f1545dc039)).

> If users want to use different API keys in different parts of code why would they wrap the whole application?

You are focused only on the Moralis case, but **the problem is general**. Let imagine, you have some API with some storage, and you want to create a migration app, like "copy files from client X to client Y" and you must do it by the SDK. The SDK must allow creating multiple instances.

image

In our case this problem will occur when:
* a user want to create a bridge app (solana -> avalanche), for example: in the beta architecture we allow to connect only one wallet per SDK instance
* the API will get any state-full resource and a user want to create a migration app

b4rtaz: One thing more, I don't want to say the current state is wrong. It's acceptable, but doesn't cover described problems. Currently those problems are edge cases so we don't need to focus on it now.

I've just saying we should have the architecture like you made (`<MoralisProvider>` + `getClient()`) because it's quite easy to fix to cover the described problems.

FedericoAmura: > > And in this case it can also break the checks that validate using hooks correctly (in renders or other hooks, no ifs, no loops)
> 
> Could you please explain it to me better?
> 
> What'd you suggest to use here?

[React Doc](https://reactjs.org/docs/hooks-rules.html)

React relies on the order the hooks are called to maintain its state between renders. That is why we cannot use them inside ifs or loops, those could change the order or times they are called.

To avoid this, they provided the ESLint plugin, which lets you know when you are writing something that breaks that rule.
Thing is, a hook is just a function whose name starts with `use...`, like `useState` or `useClient`.

By prefixing the underscore in `_useClient` you are now saying the developer that it is a "private" function while, at the same time, not saying react nor eslint that it is a hook anymore. They will not validate that rule, and will complain if you call hooks inside of it

Considering the function is not private really and is really simple, I think it is a bad tradeoff and that the `useClient` hook should be treated as a public one

Y0moo: Good point @FedericoAmura. When I used the `_` prefix I don't expect it will be "private", because it's not. It's more a tip for devs that it's not really a hook they probably need to use, but still accessible.

For now I renamed hook `_useClient` => `_useMoralisContext` and rename its property `swrConfig`=> `fetchConfig`

demos/react/src/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+import reportWebVitals from './reportWebVitals';
+import { MoralisProvider, createMoralisConfig } from '@moralisweb3/react';
+import { BrowserRouter } from 'react-router-dom';
+import { ChakraProvider } from '@chakra-ui/react';
+
+const root = ReactDOM.createRoot(document.getElementById('root') as HTMLElement);
+
+const moralisConfig = createMoralisConfig({
+  apiKey: process.env.REACT_APP_MORALIS_KEY || '',
+});
```
ErnoW, line 12: We can just provide the config, right? I don't see any reason to wrap in in a createMoralisConfig function, or is there?

ErnoW (changes requested):

demos/react/src/index.tsx
Y0moo: Yes, it's still possible to simply provide config directly to the `MoralisProvider`. Atm `createMoralisConfig` is a simple Type wrapper function that helps you to see typescript definitions when you are not providing config directly to `MoralisProvider` e.g. you have this config in separate file

Y0moo:
```
const moralisConfig = {
  apiKey: process.env.REACT_APP_MORALIS_KEY || '',
}
```
Is same to:
```
const moralisConfig = createMoralisConfig({
  apiKey: process.env.REACT_APP_MORALIS_KEY || '',
});
```
But it doesn't have type definitions. I believe it's more convenient than importing type from `MoralisProvider`

ErnoW: I see, is the more obvious choice not to just use the type like this:
```
const moralisConfig: MoralisConfig = {
  apiKey: process.env.REACT_APP_MORALIS_KEY || '',
}
```

With a function, I would expect some logic gets executed on the input to transform the data in something else. If I want type-safety then I would use a type/interface

Y0moo: Makes sense @ErnoW. Removed and updated README

ErnoW (approved):

PR #994, opened 2023-01-31
feat: client-side-support

ErnoW:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

Client-side support for:
- UMD
- Webpack


packages/common/evmUtils/src/CommonEvmUtils.ts
```diff
@@ -19,4 +19,16 @@ export class CommonEvmUtils extends Module {
...
+    return dataTypes.EvmAddress;
+  }
+
+  public get EvmChain() {
+    return dataTypes.EvmChain;
+  }
+
+  public get EvmNative() {
+    return dataTypes.EvmNative;
+  }
```
b4rtaz, line 33: I don't like this approach because those properties are completely redundant for the webpack/bundler approach. Additionally this breaks the [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle) principle (this class has already a responsibility: it's a module, but that adds a second responsibility: the UMD helper).

Maybe for the V2 this is acceptable, but for the V3 we should go into a more natural approach. I mean, we shouldn't use default exports, by this a user can use ALL types of the SDK directly from the root variable.

```html
<script src="moralis.umd.js"></script>
<script>
const client = moralis.Client();
client.evmApi.getBallance({
  address: moralis.Address.create('0x00000...')
});
</script>
```

For your consideration: this gives one benefit more. We can have almost the same syntax for UMD/bundler approaches. This should be great for tutorials. 

```ts
import * as moralis from 'moralis';

const client = moralis.Client();
client.evmApi.getBallance({
  address: moralis.Address.create('0x00000...')
});
```

demos/client-umd/scripts.js
```diff
@@ -0,0 +1,57 @@
...
+    elOut.innerHTML = 'Fetching...';
+    const { result } = await Moralis.SolApi.account.getSPL({ address });
+    elOut.innerHTML = JSON.stringify(result, null, 2);
+  } catch (error) {
+    elOut.innerHTML = JSON.stringify({ error: error.message }, null, 2);
+  }
+}
+
+// Initialise the app only when the DOM is readt to have access to the HTML elements and loaded scripts
+document.addEventListener('DOMContentLoaded', initialise);
```
b4rtaz, line 57: This comment suggests that, you shouldn't query the DOM before the DOMContentLoaded event. But at the top of this file we have:

```ts
wrapperEvmToken: document.querySelector('[data-request=evm-token]'),
wrapperSolanaToken: document.querySelector('[data-request=solana-token]'),
```

I understand this can be allowed if you attach the script after the required elements, but this comment says something opposite.

\+ typo: "readt"

demos/client-webpack/src/main.ts
```diff
@@ -0,0 +1,60 @@
...
+  try {
+    elOut.innerHTML = 'Fetching...';
+    const { result } = await Moralis.SolApi.account.getSPL({ address });
+    elOut.innerHTML = JSON.stringify(result, null, 2);
+  } catch (error: any) {
+    elOut.innerHTML = JSON.stringify({ error: error.message }, null, 2);
+  }
+}
+
+// Initialise the app only when the DOM is readt to have access to the HTML elements and loaded scripts
```
b4rtaz, line 59: Typo: `readt`. Check [this extension](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker), it helps with that kind mistakes. (Thanks @Y0moo)

b4rtaz (changes requested):

demos/client-webpack/README.md
```diff
@@ -0,0 +1,12 @@
+# Client (Webpack)
+
+This is a simple client-side implementation using a Webpack build.
+
+To get started:
+1. Copy `.env.example` to `.env` and fill in the values
+2. Run `yarn install` to install the dependencies
+3. Run `yarn build` (or `yarn build:watch`) to build the src files
+4. Run `yarn serve` to run the client on localhost
+
```
Y0moo, line 10:
```suggestion
5. You can find the demo app running on `http://192.168.1.4:7777/public`
```

Y0moo (approved):

ErnoW:
Updated the PR with:
- Validation check if Solana/Ethers/Aptos has been installed when it's getting used. Currently when using UMD, this is an external dependency, and should only be installed when the user uses these modules. Now the user gets a proper error message as soon as this package is accessed, but not installed. I don't particularly like the solution for Aptos/Solana (getAptos/getSolana), but could not find a good way to figure out if these modules have been installed otherwise, as they have a default export (when it's not installed UMD resolves it to {default:null}, otherwise it resolves to an object with all exported modules).

What do you think of this approach @Y0moo @b4rtaz , I'm not too keen on the idea that we use this getEthers/getSolana instead of importing dependencies directly, but it seems the most stable way I could find, now to handle proper validation for these modules. Alternatively we could:
- Remove this check, and document it clearly that the user needs to import these dependencies when using Solana/Aptos/Evm functionalities. But they might get weird error messages, depending on where the dependency is used. because everywhere in the code we assume that ethers/solanaweb3/aptos is defined
- Include these dependencies in the bundle, but I don't like this idea as it will blow up the bundle size

Any other thoughts?

ErnoW:
Update:
- Include ethers dependency in UMD build
- Removed Aptos and Solana SDK dependencies. We only use them for address validation, and this increases bundle size with all other code from these SDKs. As alternative I moved the code for Solana/Aptos address to our code-base

Bundle size of the UMD is pretty big though, also because of the ethers dependencies are now included. In next iteration we can find ways to limit the bundle size more, and remove some of our dependencies.

demos/client-webpack/webpack.config.js
```diff
@@ -0,0 +1,47 @@
...
+    },
+  },
+  plugins: [
+    new webpack.ProvidePlugin({
+      process: 'process/browser',
+      Buffer: ['buffer', 'Buffer'],
+    }),
+    new webpack.DefinePlugin({
+      'process.env': JSON.stringify(process.env),
+    }),
```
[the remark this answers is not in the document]
b4rtaz: `process.env` can contain sensitive information (global keys etc). So i think could be dangerous.

b4rtaz (changes requested):
Generally it looks fine.

I wonder, maybe we shouldn't have a STRICT address validation. Because this adds extra dependencies like @noble/*. and BN.js (we have own BigNumber). A simple regexp could be enough. I'm not sure.

ErnoW:
> Generally it looks fine.
> 
> I wonder, maybe we shouldn't have a STRICT address validation. Because this adds extra dependencies like @noble/*. and BN.js (we have own BigNumber). A simple regexp could be enough. I'm not sure.

I think this could be an improvement for sure. We need to know more about valid/invalid addresses in these cases. Especially if it can have different formats (like in Aptos). So I prefer an external solution that has been battle tested. I'll create a new issue to explore this a bit.

ErnoW:
> process.env can contain sensitive information (global keys etc). So I think it could be dangerous.

@b4rtaz This is client-side, so anything that you store in an env file is dangerous, as it gets included in the build. But in practice this is how apps like create-react-app / nextjs do work. That's why I consider it a non-issue.  It depends on how users use the .env file. 

Let me see a way to make it a bit more strict (like a different .env file name, adding info to Readme, or filter out only allowed env variables prefixed with REACT_), or potentially make it hardcoded.

b4rtaz:
> This is client-side, so anything that you store in an env file is dangerous

The `process.env` does not return only variables from the .env file. But from the whole OS. For example:

```js
{
  TERM_PROGRAM: 'vscode',
  PROJECT_CWD: '/Users/b4rtaz/Moralis/Moralis-JS-SDK',
  INIT_CWD: '/Users/b4rtaz/Moralis/Moralis-JS-SDK/demos/cli',
  SHELL: '/bin/zsh',
  TERM: 'xterm-256color',
  TMPDIR: '/var/folders/v5/mwqtv6w55rd02q3w_94w12x40000gn/T/',
  TERM_PROGRAM_VERSION: '1.74.3',
  MallocNanoZone: '0',
  ORIGINAL_XDG_CURRENT_DESKTOP: 'undefined',
  ZDOTDIR: '/Users/b4rtaz',
  // ...
}
```

So here I could have something sensitive. This is a reason why React has prefix for variables.

ErnoW:
> > This is client-side, so anything that you store in an env file is dangerous
> 
> The `process.env` does not return only variables from the .env file. But from the whole OS. For example:
> 
> ```js
> {
>   TERM_PROGRAM: 'vscode',
>   PROJECT_CWD: '/Users/b4rtaz/Moralis/Moralis-JS-SDK',
>   INIT_CWD: '/Users/b4rtaz/Moralis/Moralis-JS-SDK/demos/cli',
>   SHELL: '/bin/zsh',
>   TERM: 'xterm-256color',
>   TMPDIR: '/var/folders/v5/mwqtv6w55rd02q3w_94w12x40000gn/T/',
>   TERM_PROGRAM_VERSION: '1.74.3',
>   MallocNanoZone: '0',
>   ORIGINAL_XDG_CURRENT_DESKTOP: 'undefined',
>   ZDOTDIR: '/Users/b4rtaz',
>   // ...
> }
> ```
> 
> So here I could have something sensitive. This is a reason why React has prefix for variables.

Good point. I realised that we are passing in the full process.env indeed, not just values from .env. Added a filter and updated docs.

b4rtaz (approved):

demos/client-umd/README.md
```diff
@@ -0,0 +1,10 @@
+# Client (UMD)
+
+This is a simple client-side implementation using the UMD build of the library, which is imported via a `<script>` tag.
+
+To get started:
+- Replace the value of `MORALIS_API_KEY` to your own API key
```
b4rtaz, line 6: Should be there `PUBLIC_MORALIS_API_KEY`.

b4rtaz (changes requested):

b4rtaz (approved):
It's fine! 😅

Y0moo (approved):

PR #1001, opened 2023-02-03
Fix/next serialize request

Y0moo:
---
name: 'Pull request'
about: Fix @moralisweb3/next requests serialization 
---

### Checklist

### Issue Description

Originally bug is found because of "fetch function from `@moralisweb3/next` hooks throws an error when there is no chain Id param.". 

### Issue Solution

Removed unused `.serialize` and added where it's needed.


ErnoW (changes requested):
Looks good @Y0moo . Only please add a changeset

ErnoW (approved):

PR #1006, opened 2023-02-08
firebase deploy fix.

b4rtaz:
---
name: firebase deploy fix.
---


ErnoW (approved):

demos/firebase-proxy/firebase.json
```diff
@@ -4,11 +4,7 @@
     "indexes": "firestore.indexes.json"
   },
   "functions": {
-    "predeploy": [
-      "yarn --cwd \"$RESOURCE_DIR\" install",
-      "yarn --cwd \"$RESOURCE_DIR\" run lint",
-      "yarn --cwd \"$RESOURCE_DIR\" run build"
-    ]
+    "predeploy": ["cd functions && yarn install && yarn run build"]
```
Y0moo, line 7: Why everywhere we are using `yarn`? Not everyone has it and in such cases, it's better to use `npm run ...`.

b4rtaz: Because we want to work on local packages. True, by this we enforce yarn, but from out perspective we have consistent dev environment.

PR #1009, opened 2023-02-09
[1/3] create-moralis-dapp Refactor

Y0moo:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description


packages/create-moralis-dapp/next/src/template/package.json.tmpl
```diff
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "@moralisweb3/next": "^2.7.4",
-    "axios": "^1.2.1",
+    "axios": "^0.27.2",
```
b4rtaz, line 13: Why downgraded?

packages/create-moralis-dapp/next/README.md
```diff
@@ -0,0 +1 @@
+Run `create-moralis-dapp` in directory where you want to setup dApp
```
ErnoW, line 1: Should this not be `npx create-moralis-dapp`?

packages/create-moralis-dapp/toolkit/src/TemplateProcessor.ts
```diff
@@ -0,0 +1,30 @@
+import { renderFile } from 'ejs';
+import { Answers } from 'inquirer';
+import { FileSystemProcessor } from './FileSystemProcessor';
+
+export class TemplateProcessor {
+  constructor(private templatePath: string, private destination: string) {}
+
+  public async normalizeFiles(data?: Answers) {
```
ErnoW, line 8: 'normalize' suggests that the content does not change. It suggest that the format/structure only changes. Maybe renaming this to 'process' / 'generate' would be more descriptive.

packages/create-moralis-dapp/toolkit/README.md
```diff
@@ -0,0 +1 @@
+Run `create-moralis-dapp` in directory where you want to setup dApp
```
ErnoW, line 1: This is the readme of the toolkit

packages/create-moralis-dapp/toolkit/src/Inquirer.ts
```diff
@@ -0,0 +1,40 @@
+import { prompt, Answers } from 'inquirer';
+
+export type InquirerAnswers<TQuestions extends ReadonlyArray<Answers>> = Record<TQuestions[number]['name'], any>;
+export class Inquirer {
+  public static inquire<TQuestions extends ReadonlyArray<Answers>>(questions: TQuestions) {
+    return prompt<InquirerAnswers<TQuestions>>(questions);
+  }
+  public static get commonQuestions() {
```
ErnoW, line 8: It is currently setup so that we have 1 generator, that is selected by the user, and that generator is responsible for **all** questions, right?

So you slect `nextJs` and then that generator will handle all questions (including the name, apiKey etc.). Would it make more sense to handle the `common` questions in `CMDGenerator`?

Also, it seems that `confirmBeta` is already defined in `CMDGenerator` and again in the `commonQuestions`

ErnoW (changes requested):

packages/create-moralis-dapp/next/README.md
Y0moo: Both works, but will change here to `npx create-moralis-dapp`

Difference:
`npx create-moralis-dapp` will run generator and will install npm package if it's not installed.
`create-moralis-dapp` works only if it's installed on PC

packages/create-moralis-dapp/next/src/template/package.json.tmpl
Y0moo: looks like something from old files... I'll fix

packages/create-moralis-dapp/toolkit/src/Inquirer.ts
Y0moo: The current design is each generator is independent. They rely on `@create-moralis-dapp/toolkit`.

This `confirmBeta` duplication is removed on the [[3/3] PR](https://github.com/MoralisWeb3/Moralis-JS-SDK/pull/1011)

But I'll remove it here as well

ErnoW (approved):

PR #1018, opened 2023-02-21
[Streams] Fix Unindexed Event Parsing

blockframeio:
---
name: [Streams] Fix Unindexed Event Parsing
about: We discovered that log parsing for Moralis Streams was failing for an event we are tracking with an unindexed uint256[] argument.
---

### Checklist

### Issue Description

We discovered that parsing for Moralis Streams was failing for an event we are tracking with an unindexed uint256[] argument.

Related issue: N/A

### Solution Description

We fixed the issue by modifying the `LogParser` class to handle unindexed event parameters correctly. Specifically, if the param is unindexed, then we use the argument's index to access the value as opposed to the argument's name. Otherwise, the behavior is unmodified.

This fix will ensure that Moralis Streams can correctly parse events with unindexed parameters, and will prevent similar issues we faced from occurring in the future.


b4rtaz:
Could you share an example of the transaction with this problem (transactionHash)?

blockframeio:
> Could you share an example of the transaction with this problem (transactionHash)?


https://goerli.etherscan.io/tx/0x13ff9847e889923ac0b4e6a2e62508825293f744e856dddda441732e28498029

b4rtaz:
> https://goerli.etherscan.io/tx/0x13ff9847e889923ac0b4e6a2e62508825293f744e856dddda441732e28498029

Could you share the ABI for this smart contract?

b4rtaz (approved):

b4rtaz:
@blockframeio Thanks for your contribution!

PR #1031, opened 2023-03-10
@moralisweb3/react refactored with @tanstack/react-query

Y0moo:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description


demos/react/src/views/Home.tsx
```diff
@@ -34,53 +34,49 @@ const ABI = [
...
+    functionName: 'ownerOf',
+    chain: '0x1',
+    params: {
+      _tokenId: '6651',
     },
-    { revalidateOnMount: false, onSuccess: (res) => setOutput(res) },
-  );
+    abi: ABI,
+    enabled: false,
+    onSuccess: (res) => setOutput(res),
```
ErnoW, line 62: What is the benefit of combining the config with the request params? It's better to separate it:
- Better separation of concerns. One of them specifies what we send to the api, and changes per hook, the other one is configuration on handling the request, and does not change per request.
- Both definitions are not per se exclusive. We could get issues if we get an endpoint that has an `enabled` param.

ErnoW (changes requested):
I have some doubts about doing the requests only after client-side validation. If the client-side validation fails, we don't show any error, while I would expect that to be the case.

It's easy to get output that is difficult to debug for users. They provide data (but not an address), then the request shows no data, and no error. So it seems that the request is never resolved, but actually the request has never been sent. I don't think this is common practice (especially when doing api requests in React), do you have a link otherwise for reference, I'm curious to see if/how this is not common-practice in React.

We can see how the feature turns out for users, and make it otherwise configurable by the users via a config parameter, but lets make sure it is well documented.

demos/react/src/views/Home.tsx
Y0moo: These 2 designs are both commonly used in React libraries. 

Why I decided to combine them in one object:

1. It's easier to notice "callbacks/configuring" by accident when you provide parameters to the hook in an object. Really few people will try to add a second object parameter and find that it has these callbacks.
2. I agree with you, but you are thinking about the hook as a developer who maintains the code. So you think about dividing it into params sent to the Moralis API and params provided to the `react-query`. But I believe a lot of users will not think that way. All these parameters will be just `parameters` of hook for them and actually, they are.

> One of them specifies what we send to the api, and changes per hook, the other one is configuration on handling the request, and does not change per request.
3. `wagmi`  uses combined request & fetch config
4. As for me it's easier to read code:
```jsx
const { refetch: getNFTMetadata } = useEvmNFTContractMetadata({
  address: '0xbc4ca0eda7647a8ab7c2061c2e118a18a936f13d',
  chainId: '0x1',
  enabled: false,
  onSuccess: (res) => setOutput(JSON.stringify(res)),
  onError: (err) => console.log(err)
});
```
VS
```jsx
const { refetch: getNFTMetadata } = useEvmNFTContractMetadata(
  {
    address: '0xbc4ca0eda7647a8ab7c2061c2e118a18a936f13d',
    chainId: '0x1',
  },
  { 
    enabled: false,
    onSuccess: (res) => setOutput(JSON.stringify(res)),
    onError: (err) => console.log(err),
  },
);
```

In general these both designs work well for us. We need to decide only which is better.

Y0moo: We can consider renaming `enabled` to `autoFetch`.
On the other hand, imagine we will make the general config option `address` or make `wagmi` integration for extracting `address` automatically to hook requests. So users will need to have something like:

```
const { refetch: getNFTMetadata } = useEvmNFTContractMetadata(
  undefined,
  { 
    enabled: false,
    onSuccess: (res) => setOutput(JSON.stringify(res)),
    onError: (err) => console.log(err),
  },
);
```

Y0moo: Or we will have SDK operation which has only `optional` parameters. So providing first parameter object will not be mandatory. And again users will need to have this code ta have an access to callbacks/fetch config:

```
const { refetch: getNFTMetadata } = useEvmNFTContractMetadata(
  undefined,
  { 
    enabled: false,
    onSuccess: (res) => setOutput(JSON.stringify(res)),
    onError: (err) => console.log(err),
  },
);
```

ErnoW: I disagree with 'easier to read'. It's easier to read in the sense of amount of lines. But it's more difficult to grasp what are API parameters and configuration. Passing undefined might look 'ugly', but it is more explicit at least.

If we don't want to let the users pass undefined then I see 2 options:
1. Remove the api parameters from that method (and only allow passing the config)
2. Defining the options as ({request, config}), where we have 2 objects inside one 'argument-object'.
But to me having 2 arguments; `request` and `config` has the preference (or alternative 1. above)

> but you are thinking about the hook as a developer who maintains the code

Yes, and also as a developer who is using the code.
For us it is easier to maintain if we split it up. Especially if we add requests in the future that have overlap in parameter names. The current approach would need 100% a breaking change.

But also for developer it's easier to implement in my opinion. Because it is more explicit if we have 2 separate parameters.

To be fair, wagmi is the one of the only examples I have seen where they combine all arguments into 1 object. Most designs split them according to their purpose. This is especially true when we are talking about api SDKs. The `request` is what is sent to the API, and has a specific single purpose. That's probably also react-query (and similar libraries), axios and most other libraries combine all 'data' into one argument.

> It's easier to notice "callbacks/configuring" by accident when you provide parameters to the hook in an object. 

This is a small benefit, and does not outweight the cost. If they want to specify additional configurations of a mehtod they will probably:
- Check if there are additional arugments on the method
- Check the documentation
In both cases they will discover the configuration options

Y0moo: Sounds reasonable. Let's use two separate objects for request and fetch config.

Y0moo: Updated!

PR #1052, opened 2023-03-17
e2e tests.

b4rtaz:
---
name: e2e tests.
---


package.json
```diff
@@ -46,6 +46,7 @@
     "test": "yarn nx run-many --target=test --all",
     "test:watch": "yarn nx run-many --target=test:watch --all",
     "test:coverage": "yarn nx run-many --target=test:coverage --all",
+    "test:e2e": "yarn nx run-many --target=test:e2e --all",
```
ErnoW, line 49: We can have issues when multiple apps want to render at the same host localhost:3000
Adding `--parallel=1"` will solve this, but execution time is a bit longer. Or we need to have different ports for all builds.

I mainly encountered this due to the fact that one app was still running because of an incalid api key, while another app tried to load at localhost:3000

ErnoW (changes requested):

package.json
b4rtaz: Nice catch! Now the problem doesn't exist, but it appears in the future when we will have more e2e tests.

ErnoW (approved):

PR #1088, opened 2023-04-07
jsdoc for Aptos endpoint methods.

b4rtaz:
---
name: jsdoc fo endpoint methods.
---


Y0moo (dismissed):
Great job @b4rtaz! LGTM!

